PR #33, opened 2017-09-09
Ignore declarationDir compiler option

### LINKED ISSUE #32: Actual location of generated d.ts file outputs do not match expected with declarationDir
As per discussion in #30.  Bazel expects the generated d.ts files to be siblings of the generated js files.  The tsc/tsconfig declarationDir option throws this off, causing the ts_library rule to fail.

pcj:
Fixes #32

### DIFF
--- a/README.md
+++ b/README.md
@@ -78,7 +78,11 @@ Then build it:
 `bazel build //path/to/package:target`
 
 The resulting `.d.ts` file paths will be printed. Additionally, the `.js`
-outputs from TypeScript will be written to disk, next to the `.d.ts` files.
+outputs from TypeScript will be written to disk, next to the `.d.ts` files <sup>1</sup>.
+
+> <sup>1</sup> The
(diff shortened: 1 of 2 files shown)


bazel-io:

alexeagle:
How does this solution compare with just overriding the `declarationDir` in tsconfig.bzl like we do for other options like `outDir`? I'd rather treat them all the same way.

pcj:
Apologies for delay, thanks for pointing out `tsconfig.bzl`.  I force pushed a different commit.  

It fixes the issue for `ts-protoc-gen`.  In this scenario, we're always emitting a value for `declarationDir` in the generated `tsconfig.json` file, which implies that these tsconfig files will be incompatible with older versions of tsc (built before https://github.com/Microsoft/TypeScript/pull/7170, March 24, 2016).  Let me know if you think this will be an issue.

alexeagle:
Thanks!
Yes, we already specify a dependency on TS 2.4 (the tsc-wrapped compiler is tightly coupled with the version of TypeScript)

PR #37, opened 2017-09-15
emit ESmodules for es6 target

### LINKED ISSUE #21: ES6 output doesn't work
The `ts_library` rule has a `typescript.es6_sources` provider that claims to yield alternate `.closure.js` files intended to pass to bundling tools, but it doesn't actually write these outputs.

alexeagle:
Fixes #21 

/cc @gregmagolan

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -1,1 +1,2 @@
 build --strategy=TypeScriptCompile=worker
+test --test_output=errors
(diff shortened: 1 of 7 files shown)


alexeagle:
FYI, updated the commit to not bother with an aspect, as we discussed it's not needed to test for this issue. Just grabbing `typescript.es6_sources` from the direct dependencies is sufficient.

PR #38, opened 2017-09-15
Add circle CI

alexeagle:


alexeagle:
I should add a test for both worker mode and non-worker mode to this PR.
/cc @achew22

alexeagle:
The base docker image is `angular/ngcontainer` you can see where bazel gets installed in the Dockerfile: 
https://github.com/alexeagle/ngcontainer/blob/680f2b57d5262025b378155e6e4346bc7ccadc92/Dockerfile#L28
This is one of the big advantages of circleCI over travis!

alexeagle:
Now updated with both worker mode and non-worker mode.
Did this by having the user recommended settings (eg. worker mode) in the `.bazelrc`
and having a second CircleCI build that sets `--bazelrc=/dev/null`

PR #39, opened 2017-09-15
Switch devmode to UMD format.

alexeagle:
This allows faster loading in the browser (no need to transpile commonjs
modules).
We give each module a name (equivalent to the original TS sources using
the `///<amd-module name="some/name"/>` directive) so that they can be
concatenated together when served to the browser.
Also contains a transformer workaround for an upstream UMD bug in
TypeScript, which is only needed until users move to a TS release that
has the fix.

/cc @tbosch @vsavkin

### DIFF
--- a/internal/tsc_wrapped/compiler_host.ts
+++ b/internal/tsc_wrapped/compiler_host.ts
@@ -110,6 +110,20 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
   }
 
   /**
+   * TypeScript SourceFile's have a path with the rootDirs[i] still present, eg.
+   * /build/work/bazel-out/local-fastbuild/bin/path/to/file
+   * @return the path without any rootDirs, eg. path/to/file
+   */
+  private rootDirsRelative(fileName: string): string {
+    for (const root of this.options.rootDirs) {
+      if (fileName.startsWith(root)) {
+        return path.relative(root, fileName); // rootDirs are sorted longest-first
+      }
+    }
+    return fileName;
+  }
+
+  /**
    * Massages file names into valid goog.module names:
    * - resolves relative paths to the given context
    * - resolves non-relative paths which takes module_root into account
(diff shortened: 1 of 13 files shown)


internal/tsc_wrapped/umd_module_declaration_transform.ts
```diff
@@ -0,0 +1,55 @@
...
+    const visitor = (node: ts.Node): ts.Node => {
+      if (node.kind === ts.SyntaxKind.CallExpression) {
+        const ce = node as ts.CallExpression;
+        if (ce.expression.kind === ts.SyntaxKind.Identifier &&
+            (ce.expression as ts.Identifier).text === 'define' &&
+            ce.arguments.length === 2 &&
+            ce.arguments[1].kind === ts.SyntaxKind.Identifier &&
+            (ce.arguments[1] as ts.Identifier).text === 'factory') {
+          const newArguments = [ts.createLiteral(moduleNamer(sf)), ...ce.arguments];
+          ts.visitEachChild(node, visitor, context);
```
tbosch: this seems unnecessary...

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -184,12 +184,32 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
     return moduleName;
   }
 
+  // Converts file path into a valid AMD module name
+  amdModuleName(sf: ts.SourceFile) {
+    let fileName = sf.fileName;
+    // TypeScript gives us a path with the rootDirs[i] still present, eg.
+    // /build/work/bazel-out/local-fastbuild/bin/path/to/file
+    // We want path/to/file.
+    for (const root of this.options.rootDirs) {
```
tbosch: Isn't there a helper function someehre?

internal/tsc_wrapped/umd_module_declaration_transform.ts
alexeagle: oh, I see, we don't need to walk subtrees if we know there are no updates to make under there...

internal/tsc_wrapped/compiler_host.ts
alexeagle: No, right now this is duplicated with the function above that creates `goog.module` names.
I extracted a helper for it.

PR #42, opened 2017-09-20
Allow multiple tsconfig inputs to a ts_library

### LINKED ISSUE #41: tsconfig extending from other tsconfig doesn't work
We only include the tsconfig.json file the user specifies in the tsconfig attribute as inputs to a compilation. Need to also include any more tsconfig files it extends.

alexeagle:
Fixes #41


bazel-io:

PR #46, opened 2017-09-22
Turn on --strict for tsc_wrapped compilation

alexeagle:
Fix bugs it found, in particular, the umd_module_declaration_transform failed due to undefined module name

### DIFF
--- a/internal/tsc_wrapped/compiler_host.ts
+++ b/internal/tsc_wrapped/compiler_host.ts
@@ -13,6 +13,23 @@ export type ModuleResolver =
         ts.ResolvedModuleWithFailedLookupLocations;
 
 /**
+ * Narrows down the type of some properties from non-optional to required, so
+ * that we do not need to check presence before each access.
+ */
+interface BazelTsOptions extends ts.CompilerOptions {
+  rootDirs: string[];
+  rootDir: string;
+  outDir: string;
+}
+
+function narrowTsOptions(options: ts.CompilerOptions): BazelTsOptions {
+  if (!options.rootDirs) throw new Error(`compilerOptions.rootDirs should be set by tsconfig.bzl`);
+  if (!options.rootDir) throw new Error(`compilerOptions.rootDirs should be set by tsconfig.bzl`);
+  if (!options.outDir) throw new Error(`compilerOptions.rootDirs should be set by tsconfig.bzl`);
+  return options as BazelTsOptions;
+}
+
+/**
  * CompilerHost that knows how to cache parsed files to improve compile times.
  */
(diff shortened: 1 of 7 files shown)


internal/tsc_wrapped/compiler_host.ts
```diff
@@ -13,6 +13,23 @@ export type ModuleResolver =
         ts.ResolvedModuleWithFailedLookupLocations;
 
 /**
+ * Narrows down the type of some properties from non-optional to required, so
+ * that we do not need to check presence before each access.
+ */
```
evmar: This comment should be on the function, no?

package.json
```diff
@@ -4,9 +4,10 @@
     "version": "0.1.0",
     "dependencies": {
         "@types/node": "7.0.18",
-        "typescript": "2.4.x",
+        "@types/source-map": "^0.5.1",
```
alexeagle: this is now needed due to `--noImplicitAny`:
```
../../../../../../../../Projects/angular/node_modules/tsickle/built/definitions/source_map_utils.d.ts(8,93): error TS7016: Could not find a declaration file for module 'source-map'. '/usr/local/google/home/alexeagle/Projects/angular/node_modules/source-map/source-map.js' implicitly has an 'any' type.
  Try `npm install @types/source-map` if it exists or add a new declaration (.d.ts) file containing `declare module 'source-map';`
```

alexeagle: It could also be fixed if tsc-wrapped had `--skipLibCheck`

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -256,12 +275,13 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
       onError?: (message: string) => void,
       sourceFiles?: ts.SourceFile[]): void {
     // Workaround https://github.com/Microsoft/TypeScript/issues/18648
-    if (fileName.endsWith('.d.ts') && sourceFiles && sourceFiles.length > 0 &&
-        sourceFiles[0].moduleName &&
-        (this.options.module === ts.ModuleKind.AMD ||
-         this.options.module === ts.ModuleKind.UMD)) {
-      content =
-          `/// <amd-module name="${sourceFiles[0].moduleName}" />\n${content}`;
+    if ((this.options.module === ts.ModuleKind.AMD ||
```
alexeagle: just changing the order here as a micro-optimization: in google3 we should short-circuit at the first condition

internal/tsc_wrapped/compiler_host.ts
alexeagle: sure, I grabbed it from first-party, in the CL that syncs this I'll update that too

Issue #47, opened 2017-09-26
Initial setup troubles


evmar:
Following the instructions on the front page:

1.  Which version of node rules should I use?  It just says "latest" but doesn't say how to figure out what that should be.

2.  Before you run anything you need a `package.json`, otherwise you get a big error [1].

3.  That's enough to run yarn, but then after you try to use ts_library, you get another big error [2].  Alex says it's because you need a `bazel run @yarn//:yarn -- add -D @bazel/typescript`.

Error [1]:

```
$ bazel run @yarn//:yarn
ERROR: Skipping '@yarn//:yarn': no such package '@yarn//': Traceback (most recent call last):
	File "/usr/local/google/home/evanm/.cache/bazel/_bazel_evanm/f6cc45cea0037091838c829714de86cc/external/build_bazel_rules_nodejs/internal/node_install.bzl", line 100
		ctx.file("yarn.sh", ("#!/bin/bash" + "".joi...])), ...)
	File "/usr/local/google/home/evanm/.cache/bazel/_bazel_evanm/f6cc45cea0037091838c829714de86cc/external/build_bazel_rules_nodejs/internal/node_install.bzl", line 100, in ctx.file
		"".join(["\nROOT=\"$(dirname \"{}\")\"\n...])
	File "/usr/local/google/home/evanm/.cache/bazel/_bazel_evanm/f6cc45cea0037091838c829714de86cc/external/build_bazel_rules_nodejs/internal/node_install.bzl", line 100, in "".join
		"\nROOT=\"$(dirname \"{}\")\"\nNODE=\"{}\"\nSCRIPT=\"{}\"\n(cd \"$ROOT\"; \"$NODE\" \"$SCRIPT\" \"$@\")\n".format(ctx.path(package_json), ctx.path(n...), ..."))
	File "/usr/local/google/home/evanm/.cache/bazel/_bazel_evanm/f6cc45cea0037091838c829714de86cc/external/build_bazel_rules_nodejs/internal/node_install.bzl", line 106, in "\nROOT=\"$
		ctx.path(package_json)
Not a file: /usr/local/google/home/evanm/projects/devserver/package.json
WARNING: Target pattern parsing failed.
ERROR: no such package '@yarn//': Traceback (most recent call last):
	File "/usr/local/google/home/evanm/.cache/bazel/_bazel_evanm/f6cc45cea0037091838c829714de86cc/external/build_bazel_rules_nodejs/internal/node_install.bzl", line 100
[...]

PR #49, opened 2017-09-27
Add installation of rules_typescript and tsickle to the instructions

appsforartists:
Bazel will fail if you don't have `tsickle` and `rules_typescript` as `devDependencies`, so this commit adds instructions to install those.

There's a bigger question here: how come Bazel installs Node + Yarn automatically, but not its other dependencies?  Can we remove the need to manually install `rules_typescript` and `tsickle`?

See also #47.

### DIFF
--- a/README.md
+++ b/README.md
@@ -41,17 +41,19 @@ load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
 node_repositories(package_json = ["//:package.json"])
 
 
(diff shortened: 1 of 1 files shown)


appsforartists:
@googlebot is dumb.

alexeagle:
I agree the instructions need some work, thanks for these suggestions. I'm going to phrase it a bit differently.

PR #62, opened 2017-11-10
Add BUILD file for concatjs

alexeagle:


alexeagle:
I think this is red on circle because the bazel version needs updating in ngcontainer

PR #63, opened 2017-11-10
Devserver

### LINKED ISSUE #19: Add a ts_devserver rule
This requires a few things:

- Each ts_library compilation should produce the _es5_manifest.MF file that lists the topo-sorted module graph observed in this compilation. We should decompose this out of tsickle where it currently lives, and put in a TS transformer. here's a pointer: https://github.com/angular/tsickle/blob/5ac90e8499acebbe828aa1314fcf4eb86dd6b08f/src/es5processor.ts#L113 - I am hoping @lucassloan can help
(note: we do this per-library to avoid re-parsing files when we get a request)
- we need to open-source the Go implementation of our devserver and concatjs programs. Maybe @mprobst would like to get this started?
- need the Bazel rule that runs the devserver
- test that it works with ibazel

/cc @vsavkin

alexeagle:
Needs some more cleanup, depends on bazelbuild/rules_nodejs#50 to land first

/cc @gregmagolan

Fixes #19


internal/concatjs/concatjs.go
```diff
@@ -137,9 +137,10 @@ type cacheEntry struct {
 // It skips blank lines and javascript/closure/deps.js.
 // END_INTERNAL
 func manifestFiles(manifest string) ([]string, error) {
+	dir, err := os.Getwd()
 	f, err := os.Open(manifest)
 	if err != nil {
-		return nil, fmt.Errorf("could not read manifest %s: %s", manifest, err)
+		return nil, fmt.Errorf("could not read manifest %s in directory %s: %s", manifest, dir, err)
```
evmar: Isn't it always relative to the cwd?  This path is joined with whatever you passed for 'root'

internal/devserver/BUILD.bazel
```diff
@@ -0,0 +1,25 @@
+exports_files(["launcher.sh"])
+
+load("@io_bazel_rules_go//go:def.bzl", "go_binary", "go_library", "go_test")
+
+go_library(
+    name = "go_default_library",
```
evmar: Is this an autogenerated name?  Pretty weird looking.

internal/devserver/devserver.go
```diff
@@ -0,0 +1,196 @@
...
+
+type headerSuppressorResponseWriter struct {
+	http.ResponseWriter
+}
+
+// WriteHeader implements http.ResponseWriter.WriteHeader.
+func (w *headerSuppressorResponseWriter) WriteHeader(code int) {}
+
+func createFileHandler(servingPath, manifest string, pkgs []string, base string) http.HandlerFunc {
+	pkgPaths := chainedDir{}
```
evmar: In concatjs these are called 'roots', maybe we should stick to that?  "pkg" doesn't really convey that they are paths to resolve from

internal/devserver/devserver.go
```diff
@@ -0,0 +1,196 @@
...
+			return f, err
+		}
+
+		return f, nil
+	}
+	return nil, os.ErrNotExist
+}
+
+// realStatMtime returns the mtime for a file.
+func realStatMtime(filename string) (time.Time, error) {
```
evmar: unused

internal/devserver/devserver.go
```diff
@@ -0,0 +1,196 @@
...
+	s, err := os.Stat(filename)
+	if err != nil {
+		return time.Time{}, err
+	}
+	return s.ModTime(), nil
+}
+
+func realReadFile(filename string) ([]byte, error) {
```
evmar: unused

yohcop:
Hi! This is possibly a todo/issue for later, although it may be easy to add here: forward requests that either don't match any url, or that matches a given url only, to another server.
For example, to be able to forward all /api calls to a backend server.
Thanks,

alexeagle:
merged in e55b3bd2b7a68106ec7c56e7e21e1090dea90c6d

PR #70, opened 2017-11-17
Add 'yarn add' step to README

calebegg:
I had to do these additional steps to get this to work. I'm not sure if there was something else I was supposed to do, or some other way that these packages were supposed to get installed. Feel free to suggest an easier path.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,12 +54,8 @@ You can run the version Bazel has already installed:
 
 ```sh
 $ bazel run @yarn//:yarn
-```
-
-Using yarn, install this package and `tsickle`:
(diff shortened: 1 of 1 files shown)


calebegg:
I just understood that I think this comment in the readme was trying to tell me to add this. I thought it was explaining what the next line magically did somehow. 😕 . Also, even after I added `@bazel/typescript`, I still had to add `tsickle` or it failed to load that. But I don't understand why, since tsickle is in this project's package.json.

```
# Include @bazel/typescript in package.json#devDependencies
```

PR #72, opened 2017-11-17
ts_web_test rule

alexeagle:
Some things to note about this approach:
- We install a new external repo called `build_bazel_rules_typescript_karma_deps` that has all our runtime deps, so we don't mix with the users node_modules, and they don't even need to `yarn add karma`. That's good so we know we run a karma version that works with our rules.
- Doesn't yet include any way to discover and connect to an existing karma instance, so it starts a bunch of browser windows, one per test target. Can add that later

/cc @bowenni


alexeagle:
Once we like the basic shape, I'll do a copybara import into google3 and mark some files as integrated (eg. the concatjs node impl) and we can do more review internally.

internal/ts_test.bzl
```diff
@@ -0,0 +1,102 @@
...
+  # root-relative (runfiles) path to the directory containing karma.conf
+  config_segments = len(conf.short_path.split("/"))
+
+  ctx.actions.expand_template(
+      output = conf,
+      template =  ctx.file._conf_tmpl,
+      substitutions = {
+          "TMPL_runfiles_path": "/".join([".."] * config_segments),
+          "TMPL_files": TMPL_files,
+          "TMPL_workspace_name": ctx.workspace_name,
```
bowenni: I didn't find this in the template. Where is it used?

internal/ts_test.bzl
```diff
@@ -0,0 +1,102 @@
...
+      })
+
+  ctx.actions.write(
```
bowenni: Is the plan to have a karma_web_test_suite() or ts_web_test_suite() to run all of these Karma scripts?

examples/testing/BUILD.bazel
```diff
@@ -0,0 +1,20 @@
...
+)
+
+ts_library(
+    name = "tests",
+    srcs = ["decrement.spec.ts"],
+    deps = [":lib"],
+    testonly = 1,
+)
+
+ts_test(
```
bowenni: This doesn't run tests by itself right?
Which rule consumes this?

alexeagle: Yes, this does run the tests. I'll add a README about it

internal/ts_test.bzl
alexeagle: That's what I want to understand from you :)
Right now each `ts_test` just runs karma on the conf file it produces

bowenni: My original plan was that `ts_test()`s don't run by themselves because starting multiple Karma instances is expensive. I wanted to have a top-level rule/script that starts the Karma instance and shares it across all `ts_test()`s that it depends on.
But as you said "Doesn't yet include any way to discover and connect to an existing karma instance, so it starts a bunch of browser windows, one per test target. Can add that later". So I totally OK with the current way it is.

PR #73, opened 2017-11-17
Add a tsetse rule to ban == NaN

calebegg:
Based on/inspired by http://errorprone.info/bugpattern/EqualsNaN

### DIFF
--- a/internal/tsetse/checker.ts
+++ b/internal/tsetse/checker.ts
@@ -47,8 +47,18 @@ export class Checker {
    * `nodeKind` node in `nodeHandlersMap` map. After all rules register their
    * handlers, the source file AST will be traversed.
    */
+  on(nodeKind: ts.SyntaxKind.BinaryExpression,
+     handlerFunction: (checker: Checker, node: ts.BinaryExpression) => void,
+     code: number): void;
+  on(nodeKind: ts.SyntaxKind.CallExpression,
+     handlerFunction: (checker: Checker, node: ts.CallExpression) => void,
+     code: number): void;
   on(nodeKind: ts.SyntaxKind,
-     handlerFunction: (checker: Checker, node: ts.Node) => void, code: number) {
+     handlerFunction: (checker: Checker, node: ts.Node) => void,
+     code: number): void;
+  on<T extends ts.Node>(
+      nodeKind: ts.SyntaxKind,
+      handlerFunction: (checker: Checker, node: T) => void, code: number) {
     const newHandler: Handler = {handlerFunction, code};
     const registeredHandlers: Handler[]|undefined =
         this.nodeHandlersMap.get(nodeKind);
--- a/internal/tsetse/rule.ts
+++ b/internal/tsetse/rule.ts
@@ -6,8 +6,8 @@ import {Checker} from './checker';
  * files.
  */
 export abstract class AbstractRule {
-  readonly ruleName: string;
-  readonly code: number;
+  abstract readonly ruleName: string;
+  abstract readonly code: number;
 
   /**
    * Registers handler functions on nodes in Checker.
--- a/internal/tsetse/rules/equals_nan_rule.ts
+++ b/internal/tsetse/rules/equals_nan_rule.ts
@@ -1,19 +1,21 @@
 /**
- * @fileoverview Bans `== NaN` in TypeScript code, since it is always false for
- * any value.
+ * @fileoverview Bans `== NaN` and `=== NaN` in TypeScript code, since it is
+ * always false for any value.
  */
 
 import * as tsutils from 'tsutils';
 import * as ts from 'typescript';
 
 import {Checker} from '../checker';
+import {ErrorCode} from '../error_code';
 import {AbstractRule} from '../rule';
 
-const FAILURE_STRING = 'x == NaN always returns false; use isNaN(x) instead';
+const FAILURE_STRING =
+    'x == NaN and x === NaN are always false; use isNaN(x) instead';
 
 export class Rule extends AbstractRule {
   readonly ruleName = 'equals-nan';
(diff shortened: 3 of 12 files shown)


internal/tsetse/rules/equals_nan_rule.ts
```diff
@@ -0,0 +1,30 @@
...
+  register(checker: Checker) {
+    checker.on(
+        ts.SyntaxKind.BinaryExpression, checkBinaryExpression, this.code);
+  }
+}
+
+function checkBinaryExpression(checker: Checker, node: ts.BinaryExpression) {
+  if ((node.operatorToken.kind === ts.SyntaxKind.EqualsEqualsToken ||
+       node.operatorToken.kind === ts.SyntaxKind.EqualsEqualsEqualsToken) &&
+      (node.left.getText() === 'NaN' || node.right.getText() === 'NaN')) {
```
alexeagle: can we query the type checker for the actual NaN symbol rather than using the text?

I'm not sure it makes any difference though, I don't think you could get a node whose text is `NaN` other than with the real symbol.

internal/tsetse/rules/equals_nan_rule.ts
```diff
@@ -0,0 +1,30 @@
...
+
+  register(checker: Checker) {
+    checker.on(
+        ts.SyntaxKind.BinaryExpression, checkBinaryExpression, this.code);
+  }
+}
+
+function checkBinaryExpression(checker: Checker, node: ts.BinaryExpression) {
+  if ((node.operatorToken.kind === ts.SyntaxKind.EqualsEqualsToken ||
+       node.operatorToken.kind === ts.SyntaxKind.EqualsEqualsEqualsToken) &&
```
alexeagle: update comments/docs to say it covers triple-equals also

internal/tsetse/rules/equals_nan_rule.ts
```diff
@@ -0,0 +1,30 @@
...
+import * as ts from 'typescript';
+
+import {Checker} from '../checker';
+import {AbstractRule} from '../rule';
+
+const FAILURE_STRING = 'x == NaN always returns false; use isNaN(x) instead';
+
+export class Rule extends AbstractRule {
+  readonly ruleName = 'equals-nan';
+  readonly code = 21222;
```
alexeagle: hmm, we don't have a system for assigning out these numbers in a unique way. Maybe all the codes should be in one file that we import from, so we don't accidentally collide. Would you mind making that change, even though it's outside your scope?

internal/tsetse/rule.ts
```diff
@@ -6,8 +6,8 @@ import {Checker} from './checker';
  * files.
  */
 export abstract class AbstractRule {
-  readonly ruleName: string;
-  readonly code: number;
+  abstract readonly ruleName: string;
```
alexeagle: nice fix :)

internal/tsetse/rules/equals_nan_rule.ts
calebegg: According to astexplorer, TS just treats it as an identifier, not a special cased symbol.

I was surprised to see that I can actually do `var NaN = 3;` in Chrome with no error (though it doesn't actually do anything).

calebegg:
`yarn.lock` would have added those packages [when `package.json` was updated](https://github.com/bazelbuild/rules_typescript/blob/d669497a5fc7c74e607229ac88d0120c8361d321/package.json) if `yarn install` had been run. Do you want me to separate that change?

Re: review; is it easier for you if I submit changes internally or externally? I started internal but decided it might be easier to do it here.

internal/tsetse/rules/equals_nan_rule.ts
calebegg: Done. What do we want to do about internal rules? Maybe a different number space? Where/how did we get "allocated" these numbers?

internal/tsetse/rules/equals_nan_rule.ts
calebegg: Updated the `@fileoverview` and failure string. Do you think it should also be mentioned in the failure string? It seems a little verbose.

alexeagle:
Some changes are easier to do internally (depends how much testing you need under bazel vs. blaze) but since the code is open-source, this is The Right Thing To Do (IMO).
I typically develop changes in open-source so others in the community can learn from them and have a chance to comment, then pull into google3 for internal-specific review.
See instructions in `rules_typescript/copy.bara.sky` when you're ready to turn this into a CL.

alexeagle:
you didn't change package.json so you shouldn't change yarn.lock either.

internal/tsetse/checker.ts
```diff
@@ -47,8 +47,18 @@ export class Checker {
    * `nodeKind` node in `nodeHandlersMap` map. After all rules register their
    * handlers, the source file AST will be traversed.
    */
+  on(nodeKind: ts.SyntaxKind.BinaryExpression,
```
alexeagle: why these special overloads? will we need another one for every `ts.SyntaxKind`?

internal/tsetse/rules/equals_nan_rule.ts
alexeagle: yeah, I think longer is worth it to be more exact

calebegg:
Got it, thanks.

Reverted yarn.lock, I'll send a separate PR.

internal/tsetse/checker.ts
alexeagle: I'll leave that up to @bowenni

alexeagle:
now that it's green, you can mail the CL
(note, you're the first person other than me to go through this process - thanks for unknowingly volunteering :)

PR #74, opened 2017-11-19
Get JSCompiler to compile some TS code

### LINKED ISSUE #6: interoperate with rules_closure
It would be nice to have a ts_binary rule that turns ts_libraries into a single .js file.
Unless I miss something and there's already a way to do this?

Maybe with bazelbuild/rules_closure, it would be possible to optimized the output code using closure_js_binary.

achew22:
Months of teath nashing later...

Very far from perfect, but it produces a bundled JS. Launch and iterate... right?

Fixes #6


internal/common/compilation.bzl
```diff
@@ -292,6 +290,34 @@ def compile_ts(ctx,
...
+  for source in es6_sources:
+    root = source.dirname
+    es6_js_modules_roots.append(root)
+
+    root_without_prefix = root
+    prefixes_to_remove = [
+        # Bazel < 0.8.0.
+        "bazel-out/local-fastbuild/bin/",
+        # Bazel >= 0.8.0.
+        "bazel-out/k8-fastbuild/bin/",
```
codesuki: On OSX it seems to be `darwin_x86_64-fastbuild`. Maybe you could add this.

examples/closure_compiled/BUILD.bazel
```diff
@@ -0,0 +1,39 @@
...
+
+closure_js_binary(
+    name = "closure_compiled",
+    deps = [":es6_output_lib"],
+    defs = [
+        "--module_resolution=NODE",
+        "--dependency_mode=LOOSE",
+        "--js_module_root=bazel-out/k8-fastbuild/bin/examples/closure_compiled/examples/closure_compiled",
+        "--jscomp_off=reportUnknownTypes",
+        "--jscomp_off=missingSourcesWarnings",
```
mrmeku: This should be the only def you need to keep. Remove the rest

internal/common/compilation.bzl
```diff
@@ -301,6 +327,49 @@ def compile_ts(ctx,
           collect_default=True,
           collect_data=True,
       ),
+      "closure_js_library": struct(
+          # File pointing to a ClosureJsLibrary protobuf file in pbtxt format
+          # that's generated by this specific Target. It contains some metadata
```
mrmeku: Remove the unused comments?

internal/common/compilation.bzl
```diff
@@ -301,6 +327,49 @@ def compile_ts(ctx,
...
+          # prefixes. This is passed to JSCompiler via the --js_module_root
+          # flag. See find_js_module_roots() in defs.bzl.
+          js_module_roots=es6_js_modules_roots,  ### js.js_module_roots + js_module_roots,
+          # NestedSet<String> of all ES6 module name strings in the transitive
+          # closure. These are generated from the source file path relative to
+          # the longest matching root prefix. It is used to guarantee that
+          # within any given transitive closure, no namespace collisions
+          # exist. These MUST NOT begin with "/" or ".", or contain "..".
+          ###modules=js.modules + modules,
+          # NestedSet<File> of all protobuf definitions in the transitive
```
mrmeku: Remove the unused comments?

internal/common/compilation.bzl
```diff
@@ -292,6 +290,34 @@ def compile_ts(ctx,
   if not is_library:
     files += depset(tsickle_externs)
 
+  rerooted_es6_sources = []
+  es6_js_modules_roots = [] # todo(achew22): this is non recursive... recurse.
```
mrmeku: Use this code as a base to make this recursive: https://github.com/bazelbuild/rules_closure/blob/master/closure/private/defs.bzl#L61

achew22:
Hey @alexeagle. I think this works as is. Would you mind taking a look at it?

@mrmeku can you explicitly state you'd be okay with this PR being merged?

WORKSPACE
```diff
@@ -33,12 +33,22 @@ node_repositories(package_json = ["//:package.json"])
 
 http_archive(
     name = "io_bazel_rules_go",
-    url = "https://github.com/bazelbuild/rules_go/releases/download/0.7.0/rules_go-0.7.0.tar.gz",
     sha256 = "91fca9cf860a1476abdc185a5f675b641b60d3acf0596679a27b580af60bf19c",
+    url = "https://github.com/bazelbuild/rules_go/releases/download/0.7.0/rules_go-0.7.0.tar.gz",
```
mrmeku: nit: Consider reverting this part

examples/closure_compiled/BUILD.bazel
```diff
@@ -0,0 +1,40 @@
...
+    name = "parent",
+    srcs = ["main.ts"],
+    deps = [":child"],
+)
+
+closure_js_binary(
+    name = "closure_compiled",
+    deps = [":parent"],
+    defs = [
+        "--jscomp_off=reportUnknownTypes",
```
mrmeku: I think we should preemptively change this to --jscomp_off=checkTypes . I think that flag disables all the flags that are going to bite us in the future as we expand this example more.

internal/common/compilation.bzl
```diff
@@ -292,6 +291,25 @@ def compile_ts(ctx,
   if not is_library:
     files += depset(tsickle_externs)
 
+  rerooted_es6_sources = []
+  es6_js_modules_roots = []
```
mrmeku: You can remove the entire part about es6_js_module_roots and this still compiles. I think you should remove it for now because we obviously don't quite understand what this does correctly. As far as what I've read from Jart commenting on github, these roots are just stripped off from es6 import statements. I think they might only be applicable for absolute import statements (since Bazel creates a new file system)

WORKSPACE
```diff
@@ -16,8 +16,8 @@ workspace(name = "build_bazel_rules_typescript")
 
 git_repository(
     name = "build_bazel_rules_nodejs",
-    remote = "https://github.com/bazelbuild/rules_nodejs",
-    tag = "0.1.11",
+    commit = "f6eb823e1140da6eafbf77f56a249f72bb89416e",
+    remote = "https://github.com/achew22/rules_nodejs",
```
alexeagle: Do you have a PR on rules_nodejs that has the bits you need? We can't have rules_typescript on master pointing to your fork of a dependency

WORKSPACE
alexeagle: FYI I have a PR out that upgrades this to 0.7.1

examples/closure_compiled/BUILD.bazel
alexeagle: Do we need that? tsickle should suppress checkTypes on each JS file:
https://github.com/angular/tsickle/blob/8ebfd011e9386997ca94001210596c5fd75cd17d/test_files/default/default.js#L3

How do we deal with `defs` in general? I think internally at Google we've been learning towards wrapping with a macro like `ts_binary` to make things more uniform. But in open-source there is no uniformity.

/cc @evmar @rkirov

examples/closure_compiled/declarations.d.ts
```diff
@@ -0,0 +1,3 @@
+declare function someDeclaredFunction(s: string): string;
```
alexeagle: is there something in a test that asserts this isn't renamed?

internal/build_defs.bzl
```diff
@@ -132,33 +125,35 @@ def _ts_library_impl(ctx):
...
-            attr.label(allow_files = True, single_file = True),
-        "compiler":
-            attr.label(
-                default=get_tsc(),
-                single_file=False,
-                allow_files=True,
-                executable=True,
-                cfg="host"),
+        "tsconfig": attr.label(
+            allow_files = True,
```
alexeagle: curious, did you run buildifier? we should make that part of the process (maybe in a git pre-commit hook?)

WORKSPACE
yohcop: Is this right? Should it be reverted before merging?

internal/common/compilation.bzl
```diff
@@ -17,6 +17,7 @@
 
 load(":common/module_mappings.bzl", "module_mappings_aspect")
 load(":common/json_marshal.bzl", "json_marshal")
+load("@io_bazel_rules_closure//closure/private:defs.bzl", "collect_js")
```
alexeagle: does this require every user of rules_typescript to add rules_closure to their `WORKSPACE`? I don't think that's desirable, as many users will choose rollup/uglify instead. How can we avoid that?

internal/common/compilation.bzl
```diff
@@ -292,6 +291,25 @@ def compile_ts(ctx,
...
     files += depset(tsickle_externs)
 
+  rerooted_es6_sources = []
+  es6_js_modules_roots = []
+
+  for source in es6_sources:
+    root = source.dirname
+    es6_js_modules_roots.append(root)
+
+    file_path = "%s/%s" % (root, source.basename.replace(".closure", ""))
```
alexeagle: do we still produce the `foo.closure.js` outputs too?
I expected to see a change around https://github.com/bazelbuild/rules_typescript/blob/master/examples/es6_output/es6_output_test.sh#L5

examples/closure_compiled/BUILD.bazel
achew22: We may need both. The generated output definitely has suppression on checkTypes, but without this it is failing to compile ATM

mrmeku:
You can close this now Andrew, I've done all this logic in a more Alex approved way in #112

achew22:
Perfect!

PR #76, opened 2017-11-20
Interop ts_library with closure_js_binary.

mrmeku:
Every the ts_library impl now returns a struct named "closure_js_library" needed by both closure_js_library and closure_js_binary. This struct contains the srcs of the ts_library transpiled to es6 as well as the path prefixes which should be stripped off of import statements in order to find the imported file within the file system.


internal/common/compilation.bzl
```diff
@@ -329,3 +365,12 @@ def ts_providers_dict_to_struct(d):
     if type(value) == type({}):
       d[key] = struct(**value)
   return struct(**d)
+
+def collect_js(ctx, deps):
```
achew22: This is extracted from rules_closure. What would you think about moving this around in the rules_closure repo to make it an importable function?

internal/common/compilation.bzl
```diff
@@ -301,6 +329,14 @@ def compile_ts(ctx,
           collect_default=True,
           collect_data=True,
       ),
+      "closure_js_library": struct(
```
achew22: I know you like brevity, but when I'm implementing a spec, I want as much context as physically possible. Since we are implementing the rules_closure closure_js_library interface, I think we should leave the original comments along with a link to the original source.

internal/common/compilation.bzl
mrmeku: Its heavily modified from the rules_closure implementation (and the rules closure implementation was within a directory named private). Also I'm not sure that its completely wise to make rules_typescript directly depend upon rules_closure since it'll cause user's of these rules to have to clone an entire repo they won't necessarily use

internal/common/compilation.bzl
mrmeku: Donezo

WORKSPACE
```diff
@@ -16,8 +16,8 @@ workspace(name = "build_bazel_rules_typescript")
 
 git_repository(
     name = "build_bazel_rules_nodejs",
-    remote = "https://github.com/bazelbuild/rules_nodejs",
-    tag = "0.1.11",
+    commit = "f6eb823e1140da6eafbf77f56a249f72bb89416e",
+    remote = "https://github.com/achew22/rules_nodejs",
```
alexeagle: do you have a separate PR to get that change upstream?

internal/common/compilation.bzl
```diff
@@ -292,6 +290,36 @@ def compile_ts(ctx,
...
+  rerooted_es6_sources = []
+  es6_js_modules_roots = []
+  for source in es6_sources:
+    root = source.dirname
+    es6_js_modules_roots.append(root)
+
+    root_without_prefix = root
+    prefixes_to_remove = [
+        # Bazel < 0.8.0.
+        "bazel-out/local-fastbuild/bin/",
```
alexeagle: use `ctx.bin_bir.path`

internal/common/compilation.bzl
```diff
@@ -292,6 +290,36 @@ def compile_ts(ctx,
...
+    ]
+    for prefix in prefixes_to_remove:
+      if root[0:len(prefix)] == prefix:
+        root_without_prefix = root[len(prefix):]
+
+    file_path = "%s/%s" % (root_without_prefix, source.basename.replace(".closure", ""))
+    print(file_path)
+    rooted_file = ctx.actions.declare_file(file_path)
+
+    ctx.actions.run_shell(
```
alexeagle: note, we want to support Windows soon. I am not sure if this will work, it might rely on Posix `cp` command?

internal/common/compilation.bzl
alexeagle: +1 for copying in this case

internal/common/compilation.bzl
```diff
@@ -329,3 +374,12 @@ def ts_providers_dict_to_struct(d):
     if type(value) == type({}):
       d[key] = struct(**value)
   return struct(**d)
+
+def collect_js(ctx, deps):
+  """Aggregates transitive JavaScript source files from unfurled deps."""
```
alexeagle: I have no idea what it means to unfurl a dep :)

internal/common/compilation.bzl
mrmeku: Donezo

internal/common/compilation.bzl
mrmeku: Done

WORKSPACE
mrmeku: Its already merged :)

internal/common/compilation.bzl
mrmeku: Done.

internal/common/compilation.bzl
mrmeku: Removed!

mrmeku:
FYI: This PR was based off PR #74 (and includes the commit it contained) but fixes some bugs and cleans up the code a bit.

mrmeku:
We are switching back to PR #74

Issue #66, opened 2017-11-14
Document what users need to do if they bring their own devserver


alexeagle:
Things like 
- Inject the livereload script referencing `env['IBAZEL_LIVERELOAD_URL']`
- Concat the files with eval() - we have examples in node and go
- Have the ibazel tag to make sure it doesn't get restarted on changes with `ibazel run`
etc

/cc @gregmagolan

PR #79, opened 2017-11-21
 Add an example of a simple CLI.

mrmeku:
Create an example demonstrating what a server built on rules_typescript and rules_nodejs would looklike.


mrmeku:
@alexeagle I'm happy to make this into a dev server example if you think that would be beneficial! I mostly did this to demonstrate how to pull in and use external npm modules downloaded with rules_nodejs since we didn't have an example demonstrating that yet. If you think we should make this into a full dev server demo, let me know all the features you think this example should have if there are others besides the one's listed within #66

alexeagle:
I see, this is identical to the example
https://github.com/bazelbuild/rules_nodejs/tree/master/examples/rollup
except this one has the program authored in TS rather than JS.

Let's not have the example program be a devserver, it doesn't do the right thing. You could find a smaller utility library to use, and have your program just be a simple CLI.

That also makes it easier to write the e2e test for your program.

mrmeku:
Done with the implementation. I'm happy to add an end to end test if you'd like

examples/program/BUILD
```diff
@@ -0,0 +1,41 @@
...
+# Run this command via `bazel run //examples/program`
+# To pass flags to the nodejs binary use -- as a delemiter 
+# between flags for bazel and flags for the program.
+# ex. `bazel run //exmaple/program -- -C -P`
+nodejs_binary(
+    name = "program",
+    # Cause ts_library to be transpiled into ES5.
+    data = [":ts_library"],
+    # ts_library transpiles .ts files within WORKSPACE_NAME/path/to/file.js
+    entry_point = "build_bazel_rules_typescript/examples/program/main.js",
```
alexeagle: I think it's better to leave off the `.js` extension, so this also effectively references the source file

examples/program/BUILD
```diff
@@ -0,0 +1,41 @@
...
+    srcs = ["main.ts"],
+    module_name = "ts_library",
+)
+
+filegroup(name = "node_modules", srcs = glob(["node_modules/**/*"]))
+
+# Run this command via `bazel run //examples/program`
+# To pass flags to the nodejs binary use -- as a delemiter 
+# between flags for bazel and flags for the program.
+# ex. `bazel run //exmaple/program -- -C -P`
```
alexeagle: spelling - example

examples/program/BUILD
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+load("//:defs.bzl", "ts_library")
```
alexeagle: I think it's better to fully-qualify here
`@build_bazel_rules_typescript//:defs.bzl`
so it's more copy-pasteable as an example into another workspace

examples/program/BUILD
```diff
@@ -0,0 +1,41 @@
...
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+ts_library(
+    name = "ts_library",
+    srcs = ["main.ts"],
+    module_name = "ts_library",
```
alexeagle: why does it need a module_name ?

examples/program/BUILD
```diff
@@ -0,0 +1,41 @@
...
+# limitations under the License.
+
+package(default_visibility = ["//visibility:public"])
+
+load("//:defs.bzl", "ts_library")
+
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+ts_library(
+    name = "ts_library",
```
alexeagle: I'd rather not name the target the same as the rule, that's a potential source of confusion

examples/program/main.ts
```diff
@@ -0,0 +1,20 @@
+import * as program from 'commander';
```
alexeagle: somewhere we need a README about what this example is demonstrating

alexeagle:
Why closed?

PR #80, opened 2017-11-21
Update yarn.lock

calebegg:


alexeagle:
Thanks, it looks like something went wrong when merging d669497a5fc7c74e607229ac88d0120c8361d321 - we didn't get the yarn.lock file copied too.
I'll try to keep an eye out for that next time.
/cc @bowenni

PR #83, opened 2017-11-27
update CompilerHost.writeFile to support TS2.6

clydin:
Signature was slightly altered in TS 2.6.  Changes are backwards compatible with TS 2.5.

Related to angular/devkit#300
/cc @alexeagle @filipesilva


bazel-io:

clydin:
Looks like there is another in tsetse. Investigating right now.

clydin:
This should be it for 0.3.2

PR #84, opened 2017-11-27
allow tsetse to build on TS2.6

clydin:
Diagnostics in TS 2.6 are a ReadonlyArray<Diagnostic>.

Related to angular/devkit#300
/cc @alexeagle @filipesilva


bazel-io:

PR #86, opened 2017-11-28
First pass at language service plugin

calebegg:


internal/tsetse/BUILD.bazel
```diff
@@ -34,3 +34,15 @@ ts_library(
         "//internal/tsetse/rules",
     ],
 )
+
+ts_library(
+    name = "plugin",
+    srcs = [
+        "plugin.ts",
```
alexeagle: the naming is confusing, because we already export a symbol called `PLUGIN` from runner.ts.
I think these should be `language_service_plugin` or `ls_plugin` or similar

.vscode/settings.json
```diff
@@ -3,6 +3,7 @@
     "files.exclude": {
         "**/.git": true,
         "bazel-*": true
-    }
+    },
+    "typescript.tsdk": "node_modules/typescript/lib"
```
alexeagle: hum, I thought this was the default behavior in VSCode now

internal/tsetse/plugin.ts
```diff
@@ -0,0 +1,41 @@
+import * as tsForTypes from 'typescript';
+import * as tsServerLibForTypes from 'typescript/lib/tsserverlibrary';
+
+import * as perfTrace from '../tsc_wrapped/perf_trace';
```
alexeagle: I don't think perfTrace is useful in this context because we don't run the nodejs process, so we wouldn't know where to find the resulting file?

internal/tsetse/plugin.ts
```diff
@@ -0,0 +1,41 @@
...
+// editor and shows issues as semantic errors (red squiggly underline).
+
+function init() {
+  return {
+    create(info: ts.server.PluginCreateInfo) {
+      const oldService = info.languageService;
+      const checker = new Checker(oldService.getProgram());
+      // Add disabledRules to tsconfig to disable specific rules
+      // "plugins": [
+      //   {"name": "...", "disabledRules": ["equals-nan"]}
```
alexeagle: this is troubling, how would I keep that list in sync with the disabled state of the rules under `tsc`?
Seem like we should read from `bazelOpts.disabledTsetseRules` but TS doesn't let plugins see the whole tsconfig? Should we move the other config under the "plugins" section?

internal/tsetse/tsconfig.json
```diff
@@ -10,6 +10,12 @@
       "es5",
       "es2015.core",
       "es2015.collection"
+    ],
+    "plugins": [
+      // Run `bazel build internal/tsetse:plugin` for this to work.
```
alexeagle: what happens if you haven't run it?

how will users install the plugin? will there be some path under `@bazel/typescript` or do we need to publish a separate NPM package for this?

.vscode/settings.json
calebegg: It wasn't for me. Maybe I got grandfathered or something?

internal/tsetse/BUILD.bazel
calebegg: Done

internal/tsetse/plugin.ts
calebegg: Done, just blindly copy pasting :-)

internal/tsetse/plugin.ts
calebegg: TS doesn't give us direct access to the tsconfig, but of course we can just parse it ourselves -- added code to do that, wdyt? (It only runs once on startup of vscode)

(I wasn't aware of the bazelOpts thing)

internal/tsetse/tsconfig.json
calebegg: If you don't run it, it doesn't work (you get no tsetse squiggles, but everything else works).

Doesn't need to be a separate package; you can just do `@bazel/typescript/path/to/plugin.js`. But they do seem to expect you to publish it as a separate NPM module.

I don't know how NPM and TS interact; do people check in the .js? Or is there some other way to have your JS generated and then made available in NPM?

.vscode/settings.json
alexeagle: :man_shrugging:

internal/tsetse/tsconfig.json
alexeagle: Typically, you assume your users cannot compile your code, so you ship .js/.d.ts to NPM.
With Bazel, we open up the possibility of shipping sources and BUILD files. A bazel-to-bazel dependency can skip shipping .js files (`@bazel/typescript` is a source-only distro)

However, I don't think we want our users to have a broken experience before their first bazel build, so I think this plugin should probably be shipped as executable JS, which means publishing a new package. I can take on that release engineering work after this merges.

PR #87, opened 2017-11-30
Automatically serve requirejs script and livereload script when IBAZEL_LIVERELOAD_URL is set and valid

gregmagolan:


bazel-io:

internal/devserver/main/javascript.go
```diff
@@ -0,0 +1,3359 @@
+package main
+
+const RequireJs string = `/** vim: et:ts=4:sw=4:sts=4
```
alexeagle: we can't copy their code into our repo for licensing reasons.
You can use a Bazel genrule to dynamically create `javascript.go` by copying files during the build.
Even better, can we read require.js from the runfiles and serve it off the disk instead?

gregmagolan: I thought the MIT license allows that as long as the copyright header is present?

```
const RequireJs string = `/** vim: et:ts=4:sw=4:sts=4
 * @license RequireJS 2.3.5 Copyright jQuery Foundation and other contributors.
 * Released under MIT license, https://github.com/requirejs/requirejs/blob/master/LICENSE
 */
```

gregmagolan:
> how can we add an automated test for it?

Good question. Can't go in devserver_test.go. May need to the http server out of main() so it can be tested.

internal/devserver/main/javascript.go
alexeagle: Maybe so - but I have to merge this change in google3 where we have more governance rules around third-party code (for example, any product that ships with this might need an end-user-visible list of licensed software it uses, and our tooling to verify this requires a certain layout)

Also it's messy :)

gregmagolan: Hmmm. Interesting problem. So you're thinking that the ts_devserver rule would take a list of scripts to include? Would work for RequireJs although would be let user friendly to always have to pull it from npm and include it in the rule. Would cause some trouble with livereloadJS. This script is modified based on the value of IBAZEL_LIVERELOAD_URL: `livereloadScript = fmt.Sprintf(LivereloadJs, livereloadScheme, livereloadHost, livereloadPort, livereloadPort)`. I took a similar approach to https://github.com/jaschaephraim/lrserver.

alexeagle: No, I still think we should hard-code require.js is the one we want to load. But we probably should grab the version the user has installed (maybe it should be a peerDep of `@bazel/typescript`? or maybe we fix the version and make it a proper `dep` of `@bazel/typescript` and make sure we require our own version?)

gregmagolan: Ok. I think I can make it work for the livereload script as well now that I think about it. If this snippet is injected first then the vanilla livereload script should work: ```
  window.LiveReloadOptions = {
    https: true/false,
    host: theHost,
    port: thePort
  };
```

**Comment by gregmagolan** [comment]
@alexeagle Updated the PR to use `requirejs` and `livereload` peer deps. Usage looks like this: https://github.com/gregmagolan/webpack-typescript-benchmarking/tree/bazel-rxjs

Basically to you have to pull in `requirejs` and `livereload` from npm and include those folders in the `node_modules` filegroup.

Opted to generate the `javascript.go` as it seems simpler and cleaner without having to pass any file paths for scripts to the devserver executable. Works well.

If you don't add `requirejs` and `livereload` to the `node_modules` filegroup you get an error when building the devserver that looks like this: `cat: node_modules/requirejs/require.js: No such file or directory`

File: internal/devserver/main/main.go [hunk 2]
```diff
@@ -32,7 +34,26 @@ func main() {
 		os.Exit(1)
 	}
 
-	http.Handle(*servingPath, concatjs.ServeConcatenatedJS(*manifest, *base, nil /* realFileSystem */))
+	scripts := []string{RequireJs, "", ""}
```
gregmagolan: Anyone know of a better way to do this? Seems arrays are fixed length in Go so you can't push to them after declaration.

internal/devserver/main/BUILD
```diff
@@ -24,9 +24,39 @@ filegroup(
...
+   srcs = ["@//:node_modules"],
+   cmd = """
+cat > $(location :javascript.go) <<EOF
+// @filedescription DO NOT EDIT. THIS FILE IS AUTOMATICALLY GENERATED.
+package main
+
+const RequireJs string = \\`
+EOF
+
+cat node_modules/requirejs/require.js >> $(location :javascript.go)
```
gregmagolan: $(location @//:node_modules) doesn't work as its a file group. This is assuming the location is node_modules/* here.

I could something like this instead:

```
NODE_MODULES="$(locations @//:node_modules)"
```

and then grep out the locations for `require.js` and `livereload.js` from the list of strings.

gregmagolan: Updated to grep locations from $(locations @//:node_modules)

gregmagolan:
@alexeagle Instead of peer deps maybe its better to make requirejs and livereload dependencies of `rules_typescript` and have a `rules_typescript_node_modules` filegroup that is used?

gregmagolan:
@alexeagle 

Added basic node_module resolution when looking for require.js and livereload.js. These scripts are now loaded from runfiles in devserver. Also added user specified concat scripts with an `concat_files` attribute.

For user concat scripts, I opted for an attr.label_list(allow_files = True). Seemed overly complicated and unnecessary to do search through node_modules for user concat scripts when the user can specify exactly which files they want with labels or filenames.

Example usage looks like this:
```
ts_devserver(
    name = "devserver",
    deps = ["//src:app"],
    serving_path = "/bundle.js",
    static_files = ["index.html"],
    concat_files = ["user_concat.js", "node_modules/foo/bar.js"],
    tags = ["ibazel_live_reload"],
)
```

Using manifests to pass the concat_files, node_modules_files to the go binary. A manifests is used to pass the full list of node_modules files from the label list since I wasn't sure of a clean way to get the root paths from the node_modules filegroups.

Code is messy still and needs some cleanup but let me know if I'm on the right track. And should `concat_files` be named `concat_scripts` or `concat_modules` instead?

alexeagle:
I think it's good to have the user concat feature but let's not expose in the rule yet, until we understand the use case that motivates it. Anything in the rule becomes the public api and hard to change.

internal/concatjs/concatjs.go
```diff
@@ -58,6 +58,15 @@ func ServeConcatenatedJS(manifestPath, root string, fs FileSystem) http.Handler
 			w.Header().Set("Content-Encoding", "gzip")
 		}
 
+		// Write out scripts
+		for _, s := range scripts {
+			fmt.Fprint(writer, s)
+			// Ensure scripts are separated by a newline
+			if !strings.HasSuffix(s, "\n") {
```
alexeagle: why bother checking for the suffix on the string, maybe always add the newline?

internal/devserver/ts_devserver.bzl
```diff
@@ -47,6 +47,20 @@ def _ts_devserver(ctx):
   else:
     workspace_name = ctx.workspace_name
 
+  ctx.actions.write(ctx.outputs.concat_files_manifest, "".join([
+    # TODO: change to $(rootpath) after upgrading users to Bazel 0.8
```
alexeagle: maybe we should do it now

internal/devserver/ts_devserver.bzl
```diff
@@ -59,16 +73,20 @@ RUNFILES="$PWD/.."
 {main} {serving_arg} \
   -base "$RUNFILES" \
   -packages={workspace}/{package} \
-  -manifest={workspace}/{manifest}
+  -manifest={workspace}/{manifest} \
+  -concat_files_manifest={workspace}/{concat_files_manifest} \
+  -node_modules_manifest={workspace}/{node_modules_manifest}
```
alexeagle: couldn't we just give the paths directly as arguments rather than writing to a file? the string can't get longer than max argv, can it? and it makes debugging one step easier

internal/devserver/ts_devserver.bzl
```diff
@@ -84,6 +102,11 @@ ts_devserver = rule(
         "serving_path": attr.string(),
         "data": attr.label_list(allow_files = True, cfg = "data"),
         "static_files": attr.label_list(allow_files = True),
+        "concat_files": attr.label_list(allow_files = True),
+        # @// is special syntax for the "main" repository
+        # The default assumes the user specified a target "node_modules" in their
+        # root BUILD file.
+        "node_modules": attr.label_list(default = [Label("@build_bazel_rules_typescript//:node_modules"), Label("@//:node_modules")]),
```
alexeagle: it will be hard for user to override this default, because they'll need to know about their own node_modules as well as build_bazel_rules_typescript. nodejs rules have a single-valued node_modules.
So I think it would be better to keep these as two separate attributes
_rules_typescript_node_modules = attr.label(default=....
node_modules = attr.label(default=...

internal/devserver/main/main.go
```diff
@@ -32,7 +40,71 @@ func main() {
...
+	nodeModulesManifestPath := filepath.Join(*base, *node_modules_manifest)
+	node_modules_files, err := manifestFiles(nodeModulesManifestPath)
+	if err != nil {
+		fmt.Fprintf(os.Stderr, "Failed to read node_modules_manifest: %v\n", err)
+		os.Exit(1)
+	}
+
+	scripts := make([]string, 0, 100)
+
+	requireJsPath := resolveNodeModulesFile(node_modules_files, "/node_modules/requirejs/require.js");
```
alexeagle: can this go program be unaware of requirejs, and just do that as the default value for the concat_files attribute on the rule?

gregmagolan: For requirejs since devserver expects its version from package.json, I search the rules_typescript node_modules first and then the user node_modules. I don't do this for user supplied scripts in concat_files as it wouldn't make sense. So there is a difference.

internal/concatjs/concatjs.go
gregmagolan: Sure. May as well.

internal/devserver/ts_devserver.bzl
gregmagolan: The list of node_modules can be huge so having those in a file makes sense. I didn't reduce them to their base path in skylark. The list of concat_files could be changed passed as arguments although the code is simpler if its done the same way and just passed as a file. I like passing them as files as you can always look in bazel-out and see what got written out.

internal/devserver/ts_devserver.bzl
gregmagolan: Ok. I agree.

internal/devserver/ts_devserver.bzl
gregmagolan: `expand_path_into_runfiles` is in `rules_nodejs` so are you think there would be something added to `rules_nodejs` to use here that deals with `$(rootpath)`?

internal/devserver/main/main.go
alexeagle: It doesn't make sense because there is only one node_modules to resolve to, right?

internal/devserver/ts_devserver.bzl
alexeagle: we shouldn't need to use any helper from rules_nodejs, `$(rootpath)` / `ctx.expand_location` should do exactly this

internal/devserver/main/main.go
gregmagolan: No resolution for the concat_files. I opted for labels so if the user wants to use a node_modules files as a script then they would just point the exact file they want so there is no ambiguity and the build would fail if the file is not there.

gregmagolan: But yes, if concat_files was resolved then it would be just the user @//:node_modules to look through and not the rules_typescript node_modules first.

alexeagle: We can also supply labels for the requirejs dependency. If we specified two labels for that, we could have all the lookup locations we need. 
```
concat_scripts = ['$(rootpath @build_bazel_rules_typescript//:node_modules/requirejs/require.js)', '$(rootpath @//:node_modules/requirejs/require.js'] + concat_scripts
```
Then I guess the only thing would be to avoid erroring if the first one of those doesn't exist.

I keep thinking this is too complex. I think we should have rules_typescript install its own node_modules and stop relying on a layout that might be mixed with the users. https://github.com/bazelbuild/rules_nodejs/pull/67 should be a way for rules_typescript to guarantee it has a `@rules_typescript_node_modules//:all/the/files/we/expect.js`

gregmagolan: That could work for require.js. The go binary can skip files that are missing without a warning as bazel would have checked that the user supplied files exist.

I agree about having rules_typescript handle its own node_modules. You guaranteed to get what you expect where you expect it and you could also add @build_bazel_rules_typescript at a git repository instead of a local one since you're not relying on the user's npm to install the dependencies.

alexeagle: Latest thinking on this:
bazel installing `git_repository(name = "@build_bazel_rules_typescript")` means you get a source dependency on tsc_wrapped

npm installing `@bazel/typescript` should probably be a compiled distro with .js/.d.ts files, and not require the user to compile tsc_wrapped, which would cut our dependencies to not include our compile-time dependencies, like our standard best practice for npm distros

We could possibly support both, even though we know the user is using Bazel so we could always rely on source distribution.

In the latter case, is there really a need for `ts_repositories()` to install our transitive deps?

If we do have a ts_repositories, it would be nice to have `npm_install(name = "build_bazel_rules_typescript_devserver_runtime_deps")` so that users don't download these packages unless they run a `ts_devserver` target. So we would have a separate `package.json` file with the `require.js` and `livereload` packages in `devDependencies`.

In other words, we have the option to avoid mixing dependencies between `ts_library` and `ts_devserver` which seems like the idiomatic Bazel thing to do.

Let's figure this out today and get it merged.

gregmagolan: Makes sense to me

internal/devserver/package.json
```diff
@@ -0,0 +1,9 @@
+{
+    "name": "@bazel/typescript-devserver",
```
alexeagle: we won't ever publish this to npm. how about just `description` and `devDependencies`

also could you create `internal/tsc_wrapped/package.json` containing
```
{
  "description": "runtime dependencies to compile and run tsc_wrapped",
  "devDependencies": {
    "@types/node": "7.0.18",
    "@types/source-map": "^0.5.1",
    "protobufjs": "5.0.0",
    "tsickle": "0.25.x",
    "typescript": "2.5.x",
    "tsutils": "2.12.1"
  }
}
```
and point the other `npm_install` rule there?

internal/devserver/ts_devserver.bzl
```diff
@@ -47,6 +47,24 @@ def _ts_devserver(ctx):
   else:
     workspace_name = ctx.workspace_name
 
+  script_files = depset()
+  script_files += ctx.files._requirejs_script
+  script_files += ctx.files.scripts
+  ctx.actions.write(ctx.outputs.scripts_manifest, "".join([
+    # TODO: change to $(rootpath) after upgrading users to Bazel 0.8
```
alexeagle: would still like to do this now

internal/devserver/ts_devserver.bzl
```diff
@@ -47,6 +47,24 @@ def _ts_devserver(ctx):
...
+    expand_path_into_runfiles(ctx, f.path) + "\n" for f in script_files
+  ]))
+
+  livereload_script_path = expand_path_into_runfiles(ctx, ctx.files._livereload_script[0].path)
+
+  devserver_runfiles = [ctx.executable._devserver, \
```
alexeagle: why trailing backslashes?

internal/devserver/package.json
```diff
@@ -0,0 +1,9 @@
...
+    "version": "0.0.0",
+    "license": "Apache-2.0",
+    "devDependencies": {
+        "livereload": "0.6.x",
```
alexeagle: since we don't have a lockfile, let's pin these to exact versions

internal/devserver/package.json
alexeagle: with exact versions please

gregmagolan: Sure. vscode complains about missing "name" and "version" in package.json but no biggie. Npm install will probably issue a warning as well. Will check.

internal/devserver/package.json
gregmagolan: done

internal/devserver/ts_devserver.bzl
gregmagolan:
```ERROR: /private/var/tmp/_bazel_greg/5261fbd4ffdbd6b363290b88bb471f89/external/build_bazel_rules_typescript/internal/devserver/ts_devserver.bzl:62:36: syntax error at 'newline': expected expression.
ERROR: /private/var/tmp/_bazel_greg/5261fbd4ffdbd6b363290b88bb471f89/external/build_bazel_rules_typescript/internal/devserver/ts_devserver.bzl:103:1: syntax error at 'outdent': expected expression.
ERROR: error loading package '': Extension 'internal/devserver/ts_devserver.bzl' has errors.```

File: internal/devserver/package.json (hunk 9 above)
Comment (gregmagolan) [reply internal/devserver/package.json]: `@types/jasmine` required as well. I have these versions ```    "@types/jasmine": "2.8.2",
    "@types/node": "7.0.18",
    "@types/source-map": "0.5.2",
    "protobufjs": "5.0.0",
    "tsickle": "0.25.5",
    "typescript": "2.5.3",
    "tsutils": "2.12.1"
```

gregmagolan: There are newer versions available `@types/node` and `tsutils` and there is a 5.0.1 release of `protobufjs`

internal/devserver/ts_devserver.bzl
alexeagle: um, that is strange, what if you start with newline after the `[`?

gregmagolan:
@mprobst Can you review this?

alexeagle:
Martin, you'll be the G3 reviewer for the sync CL

alexeagle:
FYI this is how it looks to adopt this in a user's project:
https://github.com/alexeagle/angular-bazel-example/pull/22/files

internal/devserver/ts_devserver.bzl
gregmagolan: nope. skylark doesn't seem to like multiline expressions without `\`

gregmagolan: this works
```
  devserver_runfiles = [
    ctx.executable._devserver,
    ctx.outputs.manifest,
    ctx.outputs.scripts_manifest]
  devserver_runfiles += ctx.files.static_files
  devserver_runfiles += ctx.files._livereload_script
  devserver_runfiles += ctx.files._requirejs_script
  devserver_runfiles += ctx.files.scripts
```

gregmagolan: maybe a bit easier to read

internal/concatjs/concatjs.go
```diff
@@ -27,10 +27,10 @@ import (
 //
 // Example usage:
 //   http.Handle("/app_combined.js",
-// 	     concatjs.ServeConcatenatedJS("my/app/web_srcs.MF", ".", nil))
+// 	     concatjs.ServeConcatenatedJS("my/app/web_srcs.MF", ".", [], [], nil))
 //
 // Relative paths in the manifest are resolved relative to the path given as root.
-func ServeConcatenatedJS(manifestPath, root string, fs FileSystem) http.Handler {
+func ServeConcatenatedJS(manifestPath string, root string, preScripts []string, postScripts []string, fs FileSystem) http.Handler {
```
mprobst: Do you need both per and post scripts?

In the internal design, we just have `concatjs` pick up the `IBAZEL_...` environment variable. That's a bit more magical, but OTOH avoids having an interface dependency, and avoids having whoever calls `ServeConcatenatedJS` deal with the livereload bits.

Maybe that's easier?

internal/devserver/main/main.go
```diff
@@ -1,28 +1,37 @@
...
-	base        = flag.String("base", "", "server base (required, runfiles of the binary)")
-	pkgs        = flag.String("packages", "", "root package(s) to serve, comma-separated")
-	manifest    = flag.String("manifest", "", "sources manifest (.MF)")
-	servingPath = flag.String("serving_path", "/_/ts_scripts.js", "path to serve the combined sources at")
+	port             = flag.Int("port", 5432, "server port to listen on")
+	base             = flag.String("base", "", "server base (required, runfiles of the binary)")
+	pkgs             = flag.String("packages", "", "root package(s) to serve, comma-separated")
+	manifest         = flag.String("manifest", "", "sources manifest (.MF)")
+	scriptsManifest  = flag.String("scripts_manifest", "", "preScripts manifest (.MF)")
+	livereloadScript = flag.String("livereload_script", "", "path to live reload script")
```
mprobst: Alternative design: `IBAZEL_LIVERELOAD_SCRIPT` contains the full URL to the host serving the `.js` file that has the livereload implementation. The script can be fully self contained, including the right configuration (that you parse from the script below).

`concatjs` then just adds an IIFE loading the script via a `<script>` tag:

```javascript
(function() {
  var scriptTag = document.createElement('script');
  scriptTag.src = '${IBLAZE_LIVERELOAD_URL}';
  document.body.appendChild(scriptTag);  // or, a bit nicer, find the last `<script>` tag and append there.
}());
```

Assuming the script at the given URL then configures livereload properly (I think it even defaults to use the host it was served from, so could work with no further config), that saves a bunch of code here.

internal/devserver/main/main.go
```diff
@@ -32,7 +41,65 @@ func main() {
...
+	if match != nil && len(match) == 4 {
+		port, err := strconv.ParseUint(match[3], 10, 16)
+		if err != nil {
+			fmt.Fprintf(os.Stderr, "Cannot determine livereload port from IBAZEL_LIVERELOAD_URL")
+		} else {
+			livereloadScheme := match[1]
+			livereloadHost := match[2]
+			livereloadPort := uint16(port)
+			if *livereloadScript != "" {
+				livereloadJs, err := loadScript(filepath.Join(*base, *livereloadScript))
```
mprobst: see above, I think it's easier to have `ibazel` deal with all aspects of the live reloading, including serving the script from disk (or whereever actually).

internal/devserver/main/main.go
gregmagolan: Thanks. I think you were looking at an old version as I realized the same thing and changed it on Sat morning.

internal/concatjs/concatjs.go
gregmagolan: It could. I thought it was odd to have a concatenation aware of scripts. This way it stays generic. Also the pre and post scripts would be needed either way in the current design as requirejs is included as a preScript and the postScript is used for a the snipped `"require([\"%s\"]);", *entryModule` to start the application code.

mprobst: Right. You could move those to the bazel rules generating this if you wanted to to avoid an API dependency on that, which might give a cleaner interface. YMMV.

gregmagolan: We need a way to pass scripts to concat that are just included as-is instead of wrapping in evals and I believe requirejs needs to be included before any of the manifest files and the require([<entryModule>]) needs to be afterwards which makes need for both pre and post scripts.

gregmagolan: This API could get improved with a better manifest file that could support eval/no-eval, entry module, code splitting, etc.. @alexeagle and I have been discussing this for code splitting

mprobst:
LGTM.

PR #90, opened 2017-12-01
Use correct nodeModulesPrefix as specified by the ts_library.

### LINKED ISSUE #89: Strict dependency enforcement does not ignore .d.ts files from node_modules
If you want to reproduce the error clone my fork of rules_typescript at this commit https://github.com/mrmeku/rules_typescript/commit/a679bf70988afdaca9b933a900e0a3bd4d086309 and run 
bazel build //examples/angular:ts_library

This will produce the following error:
Compiling TypeScript (devmode) //examples/angular:ts_library failed (Exit 1).
examples/angular/main.ts(1,21): error TS2307: transitive dependency on ../../../../../../../Users/mrmeku/rules_typescript/examples/angular/node_modules/@types/angular/index.d.ts not allowed. Please add the BUILD target to your rule's deps.

I talked with Alex eagle a bit about this bug and here's what he has to say:
I added logging in an `else` case here:
https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/strict_deps.ts#L100

When we register the node_modules path with the plugin, we do a `realpath` on it to match what the TypeScript compiler should do, see comment https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/tsc_wrapped.ts#L103
but I found that the actual comparison was not using a `realpath`.

mrmeku:
Before this change node_modules was always assumed to be a subdirectory of the workspace's root.

Resolves #89

### DIFF
--- a/internal/build_defs.bzl
+++ b/internal/build_defs.bzl
@@ -87,7 +87,13 @@ def tsc_wrapped_tsconfig(ctx,
   config = create_tsconfig(ctx, files, srcs,
                            devmode_manifest=devmode_manifest,
                            **kwargs)
-  config["bazelOptions"]["nodeModulesPrefix"] = "node_modules"
+  # Default to assuming node_modules directory is located at the root of the workspace.
+  node_modules = 'node_modules'
+  if ctx.attr.node_modules.label.package:
+    node_modules = '%s/node_modules' % ctx.attr.node_modules.label.package
+
+  config["bazelOptions"]["nodeModulesPrefix"] = node_modules
+
   if config["compilerOptions"]["target"] == "es6":
     config["compilerOptions"]["module"] = "es2015"
   else:


internal/tsc_wrapped/strict_deps.ts
```diff
@@ -97,7 +94,7 @@ export function checkModuleDeps(
       // Module imports can only have one declaration location.
       const declFileName = sym.declarations[0].getSourceFile().fileName;
       if (allowedMap[stripExt(declFileName)] ||
-          (ignoredFilesPrefix && declFileName.startsWith(ignoredFilesPrefix))) {
+          declFileName.includes('/node_modules/')) {
```
mrmeku: I'm not sure this is Windows compatible. Does anyone know?

alexeagle:
I fixed it up in 9f1f3d4eb2210c8c2b93b1b07e99be58cecacc4d - does that work for you?

Thanks for working on this!

mrmeku:
Works for me! I had a similar bit of code I just forgot to commit it.
Thanks for fixing it up!

PR #91, opened 2017-12-03
Allow user to set bazelOptions.disableStrictDeps

alexeagle:
At Google, all code is required to be strict-deps compliant, except third-party code outside our control.
Users may have a similar need to opt-out some of their code, or might decide not to use this feature on
first-party code either.

Also in this commit:
- Add missing unit test for tsconfig.ts
- Include the source-map-support library so node stack traces can have .ts paths
  (also requires a change in rules_nodejs)
- Update clang-format, should (mostly?) match google-internal style
- Add --config=debug options for debugging node programs

See #89

### DIFF
--- a/internal/common/tsconfig.bzl
+++ b/internal/common/tsconfig.bzl
@@ -95,11 +95,14 @@ def create_tsconfig(ctx, files, srcs,
       # Explicitly tell the compiler which sources we're interested in (emitting
       # and type checking).
       "compilationTargetSrc": [s.path for s in srcs],
-      "disableStrictDeps": disable_strict_deps,
-      "allowedStrictDeps": [f.path for f in allowed_deps],
       "addDtsClutzAliases": getattr(ctx.attr, "add_dts_clutz_aliases", False),
       "expectedDiagnostics": getattr(ctx.attr, "expected_diagnostics", []),
   }
+
+  if disable_strict_deps:
+    bazel_options["disableStrictDeps"] = disable_strict_deps
+  else:
+    bazel_options["allowedStrictDeps"] = [f.path for f in allowed_deps]
 
   # Keep these options in sync with those in playground/playground.ts.
   compiler_options = {
--- a/internal/tsc_wrapped/BUILD.bazel
+++ b/internal/tsc_wrapped/BUILD.bazel
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 load("//:defs.bzl", "ts_library")
-load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary", "jasmine_node_test")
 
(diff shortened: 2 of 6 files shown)


WORKSPACE
```diff
@@ -17,7 +17,7 @@ workspace(name = "build_bazel_rules_typescript")
 git_repository(
     name = "build_bazel_rules_nodejs",
     remote = "https://github.com/bazelbuild/rules_nodejs",
-    tag = "0.2.2",
+    commit = "1a3a0de",
```
achew22: Please use a full SHA1

1a3a0dea51c0e53a2871fbfaba3be178a308a7a3

collisions in the reduced SHA1 address space are almost guaranteed as the commit count grows. Thanks :birthday: paradox!

BUILD.bazel
```diff
@@ -30,10 +30,19 @@ gazelle(
...
+        "tsickle",
+        "tsutils",
+        "source-map-support",
+        "protobufjs",
+        "bytebuffer",
+    ] for ext in [
+        "*.js",
+        "*.json",
+        "*.d.ts",
+    ]]),
```
evmar: Is this related to this change?

.clang-format
```diff
@@ -1 +1,2 @@
 BasedOnStyle: Google
+ColumnLimit: 100
```
evmar: Is this related to this change?

package.json
```diff
@@ -1,23 +1,28 @@
 {
     "name": "@bazel/typescript",
     "description": "Build TypeScript with Bazel",
-    "version": "0.5.0",
-    "keywords": ["typescript", "bazel"],
+    "version": "0.5.3",
```
evmar: Is this related to this change?

tools/bazel.rc
```diff
@@ -1,2 +1,5 @@
 build --strategy=TypeScriptCompile=worker
 test --test_output=errors
+
+# Enable debugging tests with --config=debug
```
evmar: Is this related to this change?

alexeagle:
I mentioned all the stuff in the commit message, so it was intentional? I can sync into separate CLs.

.clang-format
alexeagle: removed

BUILD.bazel
alexeagle: I was getting errors on Mac with exceeding max open file descriptors.
I'll make a separate change for this, or even better, figure out why bazel/ibazel have so many open FDs...

WORKSPACE
alexeagle: don't need to include this in the PR, sorry for so much noise here. Still trying to figure out how to do CI

tools/bazel.rc
alexeagle: moved to separate CL

package.json
alexeagle: no, can be separate also. Sorry this was so messy.

PR #94, opened 2017-12-03
Add skipLibCheck to default tsconfig

### LINKED ISSUE #92: Building tsc_wrapped and tsetse may require additional tsconfig options
I was going to answer on #34 but I decided to create a new issue.

Background from #34, I was having issues with tsconfig, like the OP.
I looked at the generated tsconfig, but I missed something that night. The compilation, and the AsyncIterator error wasn't happening when my code was building, it was happening when building the internal libraries of rules_typescript. So even though the generated tsconfig looked right, the build was failing (since it was failing before).

Ok, so my project uses graphql (and specifically `@types/graphql`), which is where the AsyncIterator types mentioned above are used. So graphql is in my package.json.

It seems that while any ts_library is built, all the files in node_modiles are checked?. That's a bit odd to me, but maybe typescript always does that, even if the code isn't used (actually, the repro steps below should confirm that)

So when `@bazel/typescript/internal/tsc_wrapped:perf_trace` was built the error happened. Adding "esnext" to the libs in `@bazel/typescript/internal/tsc_wrapped/tsconfig.json` was not sufficient because the ts_library for perf_trace wasn't specifying a tsconfig.

So, to make this part of the compilation work, I had to do the following:
- add `"esnext"` under lib of @bazel/typescript/internal/tsc_wrapped/tsconfig.json
- add `"esnext"` under lib of @bazel/typescript/internal/tsetse/tsconfig.json. Note that esnext is here to have typescript understand the `AsyncIterator` type (and another related one)
- Then, add `tsconfig=":tsconfig.json"` to a few libraries that needed it:
  - `@bazel/typescript/internal/tsc_wrapped:perf_trace`
  - `@bazel/typescript/internal/tsc_wrapped:plugin_api`
  - `@bazel/typescript/internal/tsetse:runner`
 
And of course, also have `esnext` in my tsconfig.json, and finally all of this work. (The build fails later, but I'm pretty confident that the above is the "right" fix for me.)

So how to fix this? I'm not quite sure. But basically as things are, the problem is that I may have a library in my package.json that requires various tsconfig options that I can't set when this package is built. And the problem seems to be that building this package requires those options as well. Maybe one option is to have this package be built with its own `package.json` and `node_modules`.

Does that make sense? If you want to reproduce:
```
git clone https://github.com/bazelbuild/rules_typescript.git
cd rules_typescript
yarn add graphql @types/graphql
bazel build examples:bar_ts_library
```

and this should produce:
```
ERROR: /tmp/rules_typescript/internal/tsc_wrapped/BUILD.bazel:62:1: Compiling TypeScript (devmode) //internal/tsc_wrapped:perf_trace failed (Exit 2)
../../../../../../../../../../tmp/rules_typescript/node_modules/@types/graphql/subscription/subscribe.d.ts(17,12): error TS2304: Cannot find name 'AsyncIterator'.
../../../../../../../../../../tmp/rules_typescript/node_modules/@types/graphql/subscription/subscribe.d.ts(29,12): error TS2304: Cannot find name 'AsyncIterable'.
Target //examples:bar_ts_library failed to build
```

If you follow the above steps (adding "esnext" to the 2 tsconfig.json files, then using tsconfig.json in all the ts_libraries, it works)

yohcop:
Fixes #92.
Also, `skipDefaultLibCheck` just above is deprecated, and should be replaced with `skipLibCheck`.

### DIFF
--- a/internal/common/tsconfig.bzl
+++ b/internal/common/tsconfig.bzl
@@ -108,10 +108,10 @@ def create_tsconfig(ctx, files, srcs,
       # Has no effect in closure/ES2015 mode. Always true just for simplicity.
       "downlevelIteration": True,
 
-      # Do not type-check the lib.*.d.ts.
+      # Do not type-check the *.d.ts.
       # We think this shouldn't be necessary but haven't figured out why yet
       # and builds are faster with the setting on.
-      "skipDefaultLibCheck": True,
+      "skipLibCheck": True,
 
       "moduleResolution": "node",
 
--- a/internal/e2e/errorchecks/erroneous.ts
+++ b/internal/e2e/errorchecks/erroneous.ts
@@ -1,1 +1,2 @@
 export const x: number = 'not a number';
+const y: TypeThatDoesNotExist = null;
--- a/internal/e2e/errorchecks/erroneous_decl.d.ts
+++ b/internal/e2e/errorchecks/erroneous_decl.d.ts
@@ -1,2 +1,0 @@
-
-export declare class Foo { field: TypeThatDoesNotExist; }


bazel-io:

yohcop:
I guess this is one way to fix #92. But maybe there are better options.

alexeagle:
I think this is obsolete?

yohcop:
Yes. The comment I have about `skipDefaultLibCheck` being deprecated still stands. Although this isn't required to fix the error I was having, it may still be good to replace `skipDefaultLibCheck` with `skipLibCheck`. See: https://www.typescriptlang.org/docs/handbook/compiler-options.html.

alexeagle:
We don't want `skipLibCheck` since that would allow syntax errors in a
.d.ts file that's a src of a ts_library.
I think we should just remove `skipDefaultLibCheck`, we need to confirm
this doesn't cause a performance regression for google internal though.

PR #97, opened 2017-12-06
Resolve the real path of node_modules

gregmagolan:
Resolve the real path of node_modules from the real exec path as compilerHost.realpath does not work from inside the bazel-sandbox. Fixes tsickle PR failure https://github.com/angular/tsickle/pull/667.
@alexeagle


bazel-io:

PR #105, opened 2017-12-14
WIP: Devserver benchmark runner

gregmagolan:
Not sure what the script should be called or which folder it should live in.

It exercised from the CI in the PR as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -33,7 +33,6 @@ jobs:
     steps:
       - checkout:
           <<: *post_checkout
-
       - restore_cache:
           key: *cache_key
       - run: bazel info release
@@ -48,7 +47,6 @@ jobs:
     steps:
       - checkout:
           <<: *post_checkout
-
       - restore_cache:
           key: *cache_key
       - run: bazel --bazelrc=/dev/null info release
@@ -62,7 +60,7 @@ jobs:
             - "node_modules"
 
   # Runs end-to-end browser tests.
-  test:
+  e2e_browser_test:
     <<: *job_defaults
     steps:
       - checkout:
@@ -74,6 +72,17 @@ jobs:
       - run: bazel run @yarn//:yarn test --script_path=yarn_test.sh
       - run: xvfb-run -a ./yarn_test.sh
 
+  # Runs devserver benchmark test.
+  devserver_benchmark_test:
+    <<: *job_defaults
+    steps:
+      - checkout:
+          <<: *post_checkout
+      - restore_cache:
+          key: *cache_key
+      - run: bazel run @yarn//:yarn
+      - run: $(bazel info bazel-bin)/external/yarn/yarn benchmark-test
+
 workflows:
   version: 2
   # Run the two builds in parallel, reporting separate status to github PRs.
@@ -81,4 +90,5 @@ workflows:
     jobs:
       - build
       - build_no_bazelrc
-      - test
+      - e2e_browser_test
+      - devserver_benchmark_test
--- a/devserver-benchmark-runner.js
+++ b/devserver-benchmark-runner.js
@@ -1,0 +1,127 @@
+const {spawn} = require('child_process');
+const fs = require('fs');
+const puppeteer = require('puppeteer');
+const tempy = require('tempy');
+
+function runBenchmark() {
+  const args = process.argv.slice(2);
+  if (args.length < 3 || args.length > 5) {
+    usage();
+    process.exit(1);
+  }
+
+  const profileFilename = tempy.file({extension: 'json'});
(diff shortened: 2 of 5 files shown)


bazel-io:

gregmagolan:
@alexeagle ^

gregmagolan:
Depends on https://github.com/bazelbuild/bazel-watcher/pull/98 abd https://github.com/bazelbuild/bazel-watcher/pull/102

gregmagolan:
CI output: https://circleci.com/gh/gregmagolan/rules_typescript/21

.circleci/config.yml
```diff
@@ -74,11 +72,23 @@ jobs:
...
+  # Runs devserver benchmark test.
+  devserver_benchmark_test:
+    <<: *job_defaults
+    steps:
+      - checkout:
+          <<: *post_checkout
+      - restore_cache:
+          key: *cache_key
+      - run: bazel run @yarn//:yarn
+      - run: $(bazel info bazel-bin)/external/yarn/yarn benchmark-test
```
gregmagolan: Not sure why but calling `bazel run @yarn//:yarn benchmark-test` hangs

alexeagle: https://github.com/alexeagle/angular-bazel-example/wiki#setup

alexeagle: `bazel run --script_path=benchmark_test.sh @yarn//:yarn benchmark-test && ./benchmark_test.sh`

devserver-benchmark-runner.js
```diff
@@ -0,0 +1,127 @@
...
+  console.log('Starting benchmark');
+  console.log(`  Devserver run target: ${runTarget}`);
+  console.log(`  File to touch: ${fileToTouch}`);
+  console.log(`  Browser url: ${browserUrl}`);
+  console.log(`  Initial timeout: ${initialTimeout}s`);
+  console.log(`  Incremental timeout: ${incrementalTimeout}s`);
+
+  console.log('Starting iBazel');
+  const child =
+      spawn('node_modules/.bin/ibazel', [`-profile_dev=${profileFilename}`, 'run', runTarget]);
```
alexeagle: we need a way to pass the `--config=ci` when the benchmark is run on CI

devserver-benchmark-runner.js
```diff
@@ -0,0 +1,127 @@
...
+  }
+
+  child.on('close', onShutdown);
+  child.on('exit', onShutdown);
+
+  // Give the initial build 5 minutes
+  waitForProfilerEvent(profileFilename, 'RUN_DONE', null, initialTimeout * 1000)
+      .then((event) => {
+        console.log('Launching chrome');
+        return puppeteer.launch()
```
alexeagle: seems cool, more compact than something like webdriver

package.json
```diff
@@ -12,20 +12,23 @@
...
+        "tempy": "^0.2.1",
         "tsickle": "0.25.x",
         "tsutils": "2.12.1",
         "typescript": "2.5.x"
     },
     "scripts": {
         "pretest": "webdriver-manager update && bazel build examples/app:all",
-        "test": "concurrently \"bazel run examples/app:devserver\" protractor --kill-others --success first"
+        "test": "concurrently \"bazel run examples/app:devserver\" protractor --kill-others --success first",
+        "benchmark-test": "node devserver-benchmark-runner.js //examples/app:devserver examples/app/app.ts http://localhost:5432 300 10"
```
alexeagle: this looks like the right shape solution to me.

How would a user adopt this? copy-paste the runner script? reach into `$(bazel info output_base)/external` to find the copy bazel already fetched? other ideas?

gregmagolan: Should just be able to run `node_modules/@bazel/typescript/devserver-benchmark-runner.js ...`

alexeagle: edited the comment - not that easy...

gregmagolan: > How would they get the needed runtime deps now that isn't an npm dependency?

Right. The recommended way to get rules_typescript is a git repo now?

> Does this need to be in the typescript layer?

Not necessarily. I suppose it could work with any devserver. Will depend on iBazel with the profiler and the devserver to have livereload. If the user has a iBazel in their path then the script could just run `ibazel` instead of `node_modules/.bin/ibazel`. Can we make that assumption?

gregmagolan: Maybe the right place for it to live is within the iBazel repo so when you pull in ibazel from npm you get the script and the required deps.

gregmagolan: Oh. But then there is the case of a non-npm ibazel install. Right.

gregmagolan:
Moving this to its own repo & npm package

PR #106, opened 2017-12-15
file_cache: don't do non-hermetic reads

alexeagle:
I had already changed the ts_library rule so that everything in node_modules is an explicit input.


alexeagle:
Note, the CI failure on bazel.io was introduced earlier - they are missing a chrome binary so the karma tests can't run.

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -359,6 +362,15 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
     return this.delegate.getDefaultLibFileName(options);
   }
 
+  realpath(s: string): string {
+    // tsc-wrapped relies on string matching of file paths for things like the
+    // file cache and for strict deps checking.
+    // TypeScript will try to resolve symlinks during module resolution which
+    // makes our checks fail.
+    // So we simply turn off symlink resolution.
```
mprobst: might want to add why we think this is safe. E.g. we have a flat node_modules tree, and so we don't usually expect to reach a single file in multiple ways?

This is also kind of annoying, the right fix on the TS side would be to use the inode...

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -343,7 +343,10 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
    */
   fileExists(filePath: string): boolean {
     // Allow moduleResolution=node to behave normally.
-    if (this.allowNonHermeticReads && isNonHermeticInput(filePath) &&
+    // TODO(alexeagle): the indexOf(node_modules) is a hack, find a better
+    // way to identify these undeclared inputs.
+    if (this.allowNonHermeticReads &&
+        filePath.split(path.sep).indexOf('node_modules') !== -1 &&
```
mprobst: Why is this needed? If everything is an input, why this?

The last time we did an indiscriminate `indexOf(node_modules)` kind of hack we caused a massive regression in google3, maybe it's better to avoid that?

alexeagle: What's happening here is I inlined the method which was removed in file_cache. It's not newly introduced.

I agree it would be better to strike any reads from non-declared inputs. I tried removing this condition, and have some typescript module resolution failures which are quite challenging. I'll keep hacking on them.

If I can't figure it out, I'd still like to merge this as it is strictly better performance than what we have today.

alexeagle: okay, I found the difference, we were passing the `files[]` array from the tsconfig to satisfy fileExists, but we should ask the cache instead.

internal/tsc_wrapped/compiler_host.ts
alexeagle: Looks like https://github.com/Microsoft/TypeScript/pull/12020 is closely related. They do symlink resolution only in this node_modules case, as you say because there are multiple ways to reach a module.
Don't we potentially have that as well? we don't always have a flat node_modules tree.

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -343,10 +343,8 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
    * file reads.
    */
   fileExists(filePath: string): boolean {
-    // Allow moduleResolution=node to behave normally.
-    if (this.allowNonHermeticReads && isNonHermeticInput(filePath) &&
-        this.delegate.fileExists(filePath)) {
-      return true;
+    if (this.allowActionInputReads) {
```
mprobst: add docs here? maybe actually expand to document the pros and cons of what we discussed? It'll be hard to swap that discussion back into your brain in three months.

alexeagle: done

PR #107, opened 2017-12-19
Produce production mode JS as a tree artifact.

alexeagle:
BREAKING CHANGE:
We no longer produce .closure.js files in Bazel. Instead, we create a path/to/label.prod directory as a
Bazel TreeArtifact (see goo.gl/tkmijs). This is similar semantics to just writing a zip file of whatever
outputs we want in production mode, without overwriting outputs from other rules.
Downstream consumers should just flatten the directories underneath the artifacts they receive.

Note: we'll probably update the es6_consumer to re-use a helper like prodmode_js_sources.bzl which provides the aspect to walk dependencies and a utility to flatten the TreeArtifacts.

Paired with @mrmeku

### DIFF
--- a/examples/es6_output/es6_output_test.sh
+++ b/examples/es6_output/es6_output_test.sh
@@ -2,17 +2,17 @@
 set -e
 
 # should not down-level ES2015 syntax, eg. `class`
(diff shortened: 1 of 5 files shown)


alexeagle:
I think this is a good idea to consider, but should be done just by modifying the `outDir` option to TypeScript, as I did for Angular's es5_esm flavor.

Anyway we don't need this PR right now, as we're copying .closure.js files out of the normal root.

PR #108, opened 2017-12-19
Expose a function which reroots production (.closure.js) files.

mrmeku:
This function is exposed so that production bundling rules
(rollup, closure, unglify etc.) can invoke this function
to get a file tree contianing only production files with
the .closure.js suffix stripped off.

### DIFF
--- a/examples/es6_output/es6_output_test.sh
+++ b/examples/es6_output/es6_output_test.sh
@@ -2,17 +2,17 @@
 set -e
 
 # should not down-level ES2015 syntax, eg. `class`
-readonly GREETER_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/greeter.closure.js)
-if [[ "$GREETER_JS" != *"class Greeter"* ]]; then
-  echo "Expected greeter.closure.js to contain 'class Greeter' but was"
-  echo "$GREETER_JS"
+readonly FOO_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/es6_output.prod/examples/foo.js)
+if [[ "$FOO_JS" != *"class Greeter"* ]]; then
+  echo "Expected foo.js to contain 'class Greeter' but was"
+  echo "$FOO_JS"
   exit 1
 fi
 
-# should have native ES Module format
-readonly MAIN_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/main.closure.js)
-if [[ "$MAIN_JS" != *"import { Greeter }"* ]]; then
-  echo "Expected main.closure.js to contain 'import { Greeter }' but was"
-  echo "$MAIN_JS"
+# should not down-level ES2015 syntax, eg. `class`
+readonly LIBRARY_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/es6_output.prod/examples/some_library/library.js)
+if [[ "$LIBRARY_JS" != *"export const cool = 1;"* ]]; then
+  echo "Expected library.js to contain 'export const cool = 1;' but was"
+  echo "$LIBRARY_JS"
   exit 1
 fi
--- a/internal/common/compilation.bzl
+++ b/internal/common/compilation.bzl
@@ -294,6 +294,11 @@ def compile_ts(ctx,
   if not is_library:
     files += depset(tsickle_externs)
 
+  transitive_es6_sources = es6_sources
+  for dep in ctx.attr.deps:
+    if hasattr(dep, "typescript"):
+      transitive_es6_sources += dep.typescript.transitive_es6_sources
+
   return {
       "files": files,
       "runfiles": ctx.runfiles(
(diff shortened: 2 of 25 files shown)


examples/es6_output/es6_output_test.sh
```diff
@@ -2,15 +2,15 @@
 set -e
 
 # should not down-level ES2015 syntax, eg. `class`
-readonly GREETER_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/greeter.closure.js)
+readonly GREETER_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/es6_output.prod/examples/es6_output/greeter/greeter.js)
 if [[ "$GREETER_JS" != *"class Greeter"* ]]; then
   echo "Expected greeter.closure.js to contain 'class Greeter' but was"
```
alexeagle: update text

examples/es6_output/es6_output_test.sh
```diff
@@ -2,15 +2,15 @@
...
   echo "$GREETER_JS"
   exit 1
 fi
 
 # should have native ES Module format
-readonly MAIN_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/main.closure.js)
+readonly MAIN_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/es6_output.prod/examples/es6_output/main.js)
 if [[ "$MAIN_JS" != *"import { Greeter }"* ]]; then
   echo "Expected main.closure.js to contain 'import { Greeter }' but was"
```
alexeagle: same

examples/es6_output/es6_output_test.sh
```diff
@@ -2,15 +2,15 @@
 set -e
 
 # should not down-level ES2015 syntax, eg. `class`
-readonly GREETER_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/greeter.closure.js)
+readonly GREETER_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/es6_output/es6_output.prod/examples/es6_output/greeter/greeter.js)
```
alexeagle: shouldn't this be `es6_output_lib.prod` ?

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -130,8 +130,8 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
 
   /** Allows suppressing warnings for specific known libraries */
   shouldIgnoreWarningsForPath(filePath: string): boolean {
-    return this.bazelOpts.ignoreWarningPaths.some(
-        p => !!filePath.match(new RegExp(p)));
+    return (this.bazelOpts.ignoreWarningPaths || [])
```
alexeagle: is this change required for it to work?

internal/common/compilation.bzl
```diff
@@ -292,6 +290,10 @@ def compile_ts(ctx,
   if not is_library:
     files += depset(tsickle_externs)
 
+  transitive_es6_srcs = depset()
+  for dep in ctx.attr.deps:
+    transitive_es6_srcs += getattr(dep.typescript, "transitive_es6_srcs", [])
```
alexeagle: I think it would be better to do this with an aspect, maybe we should pair on that?

internal/common/reroot_prod_files.bzl
```diff
@@ -0,0 +1,24 @@
...
+  rerooted_es6_srcs = depset()
+
+  for dep in ctx.attr.deps:
+    if hasattr(dep, "typescript"):
+      for es6_src in dep.typescript.transitive_es6_srcs:
+        rerooted_es6_src = ctx.actions.declare_file(
+          "%s.prod/%s" % (
+            ctx.label.name,
+            es6_src.short_path.replace(".closure.js", ".js")))
+        ctx.actions.expand_template(
```
alexeagle: comment that this is a cheap way of doing a file copy?

examples/es6_output/es6_output_test.sh
mrmeku: No, we use the label of the bundling rule rather than each individual ts_library since we are rerooting under a single tree rather than many.

examples/es6_output/es6_output_test.sh
mrmeku: Donezo

examples/es6_output/es6_output_test.sh
mrmeku: Donezo

internal/common/compilation.bzl
mrmeku: I did it myself! Watching you do it yesterday was all I needed

internal/tsc_wrapped/compiler_host.ts
mrmeku: No, but it is a bug in rules_typescript that we might as well merge in. I ran into it frequently while doing ABC work

alexeagle: hmm, I'd rather keep it separate, and also understand why this is sometimes not set (I think it's easier to understand if the data is not nullable)

mrmeku:
Test this please

thelgevold:
How can I get access to this function from my external application? I might be misunderstanding this, but I thought the function had to be added to `defs.bzl` for me to be able to import it?

mrmeku:
This was added to rules_nodejs instead

PR #113, opened 2017-12-20
Add skylint to circleci

alexeagle:


alexeagle:
It's broken even on a "large" machine.

Maybe @gregmagolan has some time to add skylint to the docker image (note, I had problems doing that which might need upgrade to jdk9 to resolve)

PR #115, opened 2017-12-30
Allow ts_library to contain only .d.ts files

### LINKED ISSUE #24: ts_library fails when only composed of .d.ts files
You get the following error. As a temporary fix you can just include an empty `.ts` file

```
ERROR: /home/achew22/js/externs/BUILD:3:1: in ts_library rule //externs:externs: 
Traceback (most recent call last):
        File "/home/achew22/js/externs/BUILD", line 3
                ts_library(name = 'externs')
        File "/home/achew22/.cache/bazel/_bazel_achew22/b4046dd61443a6da2a0d424d121b9b4a/external/build_bazel_rules_typescript/internal/build_defs.bzl", line 116, in _ts_library_impl
                compile_ts(ctx, is_library = True, compile_acti..., <2 more arguments>)
        File "/home/achew22/.cache/bazel/_bazel_achew22/b4046dd61443a6da2a0d424d121b9b4a/external/build_bazel_rules_typescript/internal/common/compilation.bzl", line 199, in compile_ts                compile_action(ctx, inputs, outputs, tsconfig_json_es...)
        File "/home/achew22/.cache/bazel/_bazel_achew22/b4046dd61443a6da2a0d424d121b9b4a/external/build_bazel_rules_typescript/internal/build_defs.bzl", line 53, in compile_action
                ctx.action(progress_message = "Compiling Ty..., <6 more arguments>)
illegal argument in call to action: action outputs may not be empty.
```

mrmeku:
Fixes #24 
Previously having such a ts_library would error out because the es5 compilation step would not have any outputs (which makes sense since only the es6 compilation step produces the .externs file)

### DIFF
--- a/examples/BUILD.bazel
+++ b/examples/BUILD.bazel
@@ -17,8 +17,14 @@ package(default_visibility = ["//visibility:public"])
 load("//:defs.bzl", "ts_library")
 
 ts_library(
+    name = "declarations",
+    srcs = ["declarations.d.ts"],
+)
+
+ts_library(
     name = "foo_ts_library",
     srcs = [
+        ":declarations",
         "foo.ts",
         "types.d.ts",
     ],
--- a/examples/declarations.d.ts
+++ b/examples/declarations.d.ts
@@ -1,0 +1,1 @@
+declare const globalVariable = 'globalVariable';
--- a/examples/foo.ts
+++ b/examples/foo.ts
@@ -24,5 +24,7 @@ export class Greeter implements Polite {
   }
 }
 
+console.log(globalVariable);
+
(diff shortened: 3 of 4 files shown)


alexeagle:
Something we need to figure out here is, what guarantees should there be against a syntax/semantic error in a .d.ts file? This PR just turns off `tsc` altogether which means we stop checking them.

mrmeku:
Alex are you sure they aren't being processed? I'm pretty sure they are processed during the es6 transpilation step. This change just disables doing anything during the es5 transpilation step since there's nothing that needs to be done for es5 mode. Is your concern that if the es6 sources aren't requested that the .d.ts files won't be checked?

PR #129, opened 2018-01-17
Clarify role of rules_nodejs in rules_typescript installation

pshields:
I noticed the README did not explain why rules_nodejs is needed to use rules_typescript. I've attempted to provide a concise explanation. This stemmed from my attempt to better understand the boundaries between what rules_typescript and rules_nodejs are responsible for.

I also took this chance to update the rules_nodejs and rules_go releases referenced in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,10 +32,13 @@ containing:
 ```python
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# rules_nodejs provides build rules related to NPM and Yarn package management.
+# These rules are used for modeling the dependencies between user TypeScript
+# code and NPM or Yarn packages.
 git_repository(
     name = "build_bazel_rules_nodejs",
     remote = "https://github.com/bazelbuild/rules_nodejs.git",
-    tag = "0.0.2", # check for the latest tag when you install
+    tag = "0.3.1", # check for the latest tag when you install
 )
 
 load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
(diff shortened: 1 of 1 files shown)


bazel-io:

README.md
```diff
@@ -32,10 +32,13 @@ containing:
...
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# rules_nodejs provides build rules related to NPM and Yarn package management.
+# These rules are used for modeling the dependencies between user TypeScript
+# code and NPM or Yarn packages.
 git_repository(
     name = "build_bazel_rules_nodejs",
     remote = "https://github.com/bazelbuild/rules_nodejs.git",
-    tag = "0.0.2", # check for the latest tag when you install
+    tag = "0.3.1", # check for the latest tag when you install
```
yohcop: I'd clarify here too: where to check (i.e. the bazelbuild/rules_typescript/WORKSPACE file correct? This has to match what's in this project's WORKSPACE as I understand?)

pshields: I believe the current instructions mean to pick the latest tag from https://github.com/bazelbuild/rules_nodejs/releases, irrespective of what's in rules_typescript's WORKSPACE. I wasn't under the impression the versions of these repos need to match across workspaces.

I've clarified the comment to say where to check for the latest tag. Can one of the admins verify my interpretation?

alexeagle: it's possible for the latest rules_nodejs to contain a breaking change, where version x of rules_typescript doesn't work with it. However, the latest version of rules_typescript *ought* to always work with rules_nodejs. Sadly WORKSPACE doesn't have the peerDependency or dependency range semantics of npm that allow rules_typescript to indicate the range of versions of rules_nodejs that work.

For this reason, I plan to remove the versioning of rules from WORKSPACE to the extent possible, where a package.json could be used instead.

I think renovate is the answer here - if I install it on this repo, it will keep the version of rules_nodejs up-to-date in the WORKSPACE, along with CI that ensures it works. Then the correct guidance is to copy the rules_nodejs version from the rules_typescript WORKSPACE (manually emulating the missing transitive dependencies in Bazel)

README.md
```diff
@@ -32,10 +32,14 @@ containing:
 ```python
 load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
 
+# rules_nodejs provides build rules related to NPM and Yarn package management.
```
alexeagle: actually, rules_nodejs is more than this - any rules which are not typescript-specific go there. for example, there is `jasmine_node_test` and soon we'll merge `rollup_bundle`.

pshields: Thanks - updated. I decided to still lead with the package management part since that seems like the primary use case.

pshields:
Looks like this is stale now.

Issue #134, opened 2018-01-24
Cannot find module 'tmp' in ts_web_test


yohcop:
Sorry I don't have right now a simple repository demonstrating this error. My project is rather large, and it's difficult to reproduce.

Note that my project does not depend on the package `'tmp'`. The 'tmp' missing module is needed by this repo, and used in `internal/karma/index.ts`.

For rules_typescript, I am using: 
```
    "@bazel/typescript": "git://github.com/bazelbuild/rules_typescript.git#0.9.0",
```
in my package.json file, and for rules_nodejs, I have `commit = 609df7cb93ab412acabc40ead193cbe945dda674` in my WORKSPACE (same as the WORKSPACE file in this repo).

Here is the output of a test command. I know this is not much. Hopefully somebody has an idea of what could be going wrong. Otherwise, I'll try to reproduce in a smaller repository.

```
> bazel test fitzenith/web/app/users:testing
INFO: Analysed target //fitzenith/web/app/users:testing (46 packages loaded).
INFO: Found 1 test target...
ERROR: /home/y/.cache/bazel/_bazel_y/353cc7731d9fb1ef948c81259e5d7654/external/build_bazel_rules_typescript/internal/karma/BUILD.bazel:7:1: Compiling TypeScript (devmode) @build_bazel_rules_typescript
//internal/karma:karma_concat_js failed (Exit 1)
external/build_bazel_rules_typescript/internal/karma/index.ts(8,22): error TS2307: Cannot find module 'tmp'.

Target //fitzenith/web/app/users:testing failed to build
Use --verbose_failures to see the command lines of failed build steps.
INFO: Elapsed time: 21.555s, Critical Path: 15.84s
FAILED: Build did NOT complete successfully

Executed 0 out of 1 test: 1 fails to build.
```

alexeagle:
Yeah, this is a bug, I observed it also.

yohcop:
Uh oh. Now (after upgrading from 0.7.1 to 0.9.0), I see it even on build:

```
> bazel build fitzenith/web/app/...
INFO: Analysed 80 targets (1 packages loaded).
INFO: Found 80 targets...
[...]

PR #137, opened 2018-01-26
Don't use types for 'tmp' in karma/index.ts

gregmagolan:
Fix for https://github.com/bazelbuild/rules_typescript/issues/134


bazel-io:

alexeagle:
This fix masks an actual bug, so I'm going to make the real fix

PR #138, opened 2018-01-27
Include .js files as inputs to ts_library actions.

alexeagle:
TS module resolution sometimes follows the main field in package.json to resolve them.
Also give explicit node_modules for the karma concat-js plugin instead of relying on the user.

### DIFF
--- a/internal/build_defs.bzl
+++ b/internal/build_defs.bzl
@@ -42,7 +42,7 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file):
     return struct()
 
   action_inputs = inputs + [f for f in ctx.files.node_modules + ctx.files._tsc_wrapped_deps
-                            if f.path.endswith(".ts") or f.path.endswith(".json")]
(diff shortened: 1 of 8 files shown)


alexeagle:
This is the output from `tsc --traceResolution` that led me to find the bug:

```
Found 'package.json' at '/private/var/tmp/_bazel_alexeagle/a09cf724fa54eb41a4db0371180a1b62/execroot/angular_bazel_example/external/build_bazel_rules_typescript_deps/node_modules/tmp/package.json
'package.json' has 'main' field 'lib/tmp.js' that references '/private/var/tmp/_bazel_alexeagle/a09cf724fa54eb41a4db0371180a1b62/execroot/angular_bazel_example/external/build_bazel_rules_typescript_deps/node_modules/tmp/lib/tmp.js'.
File  '/private/var/tmp/_bazel_alexeagle/a09cf724fa54eb41a4db0371180a1b62/execroot/angular_bazel_example/external/build_bazel_rules_typescript_deps/node_modules/tmp/lib/tmp.js' does not exist.
...
======== Module name 'tmp' was not resolved. ========
```

PR #139, opened 2018-01-28
WIP: typeRoots support in ts_library

gregmagolan:


bazel-io:

gregmagolan:
There is an easier way to accomplish the same thing with the existing rule.

PR #147, opened 2018-02-09
update README.md to more recent versioned exmaples

Toxicable:
The current example will result in an error.
This change locks the versions so the example will always work.


bazel-io:

PR #149, opened 2018-02-15
Produce a nodejs profile together with a perf_trace.

mprobst:
This allows source level profiling of the JavaScript code, in addition to the more coarsely grained application level profile.


bazel-io:

Issue #152, opened 2018-02-19
Fail to compile tsetse due to overwriting outputs


achew22:
Running rules_typescript from inside a container is failing @ `d3cc5cd72d89aee0e4c2553ae1b99c707ecbef4e` (current HEAD as of posting).

Inside my docker container when I run `bazel build @build_bazel_rules_typescript//internal/tsetse/rules:rules` (not discovered by running it directly, I tried to build my target and this was a dep) I get:

```
ERROR: /var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/external/build_bazel_rules_typescript/internal/tsetse/rules/BUILD:19:1: Compiling TypeScript (devmode) @build_bazel_rules_typescript//internal/tsetse/rules:rules failed (Exit 1)
error TS5033: Could not write file '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/checker.js': EACCES: permission denied, open '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/checker.js'.
error TS5033: Could not write file '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/error_code.js': EACCES: permission denied, open '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/error_code.js'.
error TS5033: Could not write file '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/failure.js': EACCES: permission denied, open '/var/jenkins_home/.cache/bazel/_bazel_jenkins/012289bd6d606b886bac5c41b583ff44/execroot/vics/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsetse/failure.js'.
[...]

PR #158, opened 2018-03-01
Windows fixes

gregmagolan:
/cc @alexeagle


gregmagolan:
Curated and improved fixes from #155

WORKSPACE
```diff
@@ -14,15 +14,14 @@
 
 workspace(name = "build_bazel_rules_typescript")
 
-# Using a pre-release snapshot to pick up a commit that makes all nodejs_binary
-# programs produce source-mapped stack traces.
-RULES_NODEJS_VERSION = "926349cea4cd360afcd5647ccdd09d2d2fb471aa"
+# Using a pre-release snapshot to pick up a fix for puppeteer
+RULES_NODEJS_VERSION = "0162fdbe8ed986c9b5d5b79e53c98385ddaf6edd"
```
alexeagle: don't think you need this anymore?

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,59 @@
 // Karma configuration
 // GENERATED BY Bazel
+const path = require('path');
+const fs = require('fs');
+const tmp = require('tmp');
+
+process.env.CHROME_BIN = require('puppeteer').executablePath();
```
alexeagle: in theory the `ts_web_test` should allow the user to select one or more browsers,
https://github.com/bazelbuild/rules_typescript/blob/master/internal/karma/ts_web_test.bzl#L22

just to keep in mind, that anything hard-coded will need to be revisited

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,59 @@
...
+    });
+  const manifestKeys = Object.keys(manifest);
+  files = files.map((f) => {
+    if (manifestKeys.includes(f)) {
+      return manifest[f];
+    } else {
+      throw new Error(`File not found in MANIFEST: ${f}`);
+    }
+  });
+} else {
```
alexeagle: do we need to support a case where there is no `MANIFEST`?

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,59 @@
...
+  files = files.map((f) => f.replace('build_bazel_rules_typescript/external/', ''));
+}
+
+let requireFiles = [
+  TMPL_files
+].map((f) => f.replace('build_bazel_rules_typescript/external/', ''));
```
alexeagle: why is rules_typescript special here? Could there be other external repos in the files?

internal/karma/karma.conf.js
```diff
@@ -49,8 +103,8 @@ module.exports = function(config) {
 
     // base path that will be used to resolve all patterns (eg. files, exclude)
     basePath: 'TMPL_runfiles_path',
-    files: [
-TMPL_files
-    ]
+
+    // files
```
alexeagle: stray comment?

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -7,6 +7,7 @@ import {FileLoader} from './file_cache';
 import * as perfTrace from './perf_trace';
 import {BazelOptions} from './tsconfig';
 import {DEBUG, debug} from './worker';
+import {relative, join} from './path';
```
alexeagle: how about `import * as path from './path'` so you don't need other changes below

internal/tsc_wrapped/path.ts
```diff
@@ -0,0 +1,21 @@
+import * as path from 'path';
+
+// resolve function that maintains backslashes
```
alexeagle: 'maintains' seems like the wrong word for these? do you mean "maintains forward slashes"?

I'm curious if we can find code inside the TypeScript compiler that does approx. the same thing, which would validate this.

They have compilerHost.canonicalFilename for this purpose, maybe we should actually be calling that instead of manipulating the slashes?

internal/tsc_wrapped/strict_deps.ts
```diff
@@ -17,6 +17,7 @@
 
 import * as path from 'path';
 import * as ts from 'typescript';
+import {relative} from './path';
```
alexeagle: again, `import * as path from 'path'`

internal/tsc_wrapped/tsc_wrapped.ts
```diff
@@ -12,6 +11,7 @@ import {PLUGIN as strictDepsPlugin} from './strict_deps';
 import {BazelOptions, parseTsconfig} from './tsconfig';
 import {fixUmdModuleDeclarations} from './umd_module_declaration_transform';
 import {debug, log, runAsWorker, runWorkerLoop} from './worker';
+import {resolve} from './path';
```
alexeagle: again

internal/tsc_wrapped/tsconfig.ts
```diff
@@ -17,6 +17,7 @@
 
 import * as path from 'path';
 import * as ts from 'typescript';
+import {resolve, normalize} from './path';
```
alexeagle: again

internal/tsc_wrapped/tsconfig_test.ts
```diff
@@ -15,10 +15,10 @@
  * limitations under the License.
  */
 
-import * as path from 'path';
 import * as ts from 'typescript';
 
 import {parseTsconfig} from './tsconfig';
+import {resolve} from './path';
```
alexeagle: again

internal/tsc_wrapped/worker.ts
```diff
@@ -1,4 +1,4 @@
-import * as path from 'path';
+import {resolve} from './path';
```
alexeagle: again

WORKSPACE
gregmagolan: you mean switch to git_repository from the skylark rules?

internal/karma/karma.conf.js
gregmagolan: I'll add a comment to that line explaining it a bit. Karma will just look for CHROME_BIN in the environment if it is configured to use Chrome. If another browser is selected then that value is not used.

internal/karma/karma.conf.js
gregmagolan: that is to handle the two files from ts_web_test.bzl: ```  files_entries += [
    "build_bazel_rules_typescript/external/build_bazel_rules_typescript_karma_deps/node_modules/requirejs/require.js",
    "build_bazel_rules_typescript/external/build_bazel_rules_typescript_karma_deps/node_modules/karma-requirejs/lib/adapter.js",
  ]
```
but yes, I suppose there could be other externals included so should change the replace to handle any external workspace

internal/karma/karma.conf.js
gregmagolan: same as comment below. `build_bazel_rules_typescript/external/` prefix is what is needed to match in the manifest and that needs to be dropped so it matches on runfiles.

internal/karma/karma.conf.js
gregmagolan: yeah. a bit redundant. there was a comment before every other property and I didn't know what else to put.

internal/tsc_wrapped/strict_deps.ts
gregmagolan: this is importing from the local `./path.ts` which provides versions of the four path functions which maintain backslashes. you think it would be better as `import * as path from './path'`? may be a bit confusing.

internal/karma/karma.conf.js
gregmagolan: I can add a comment here that explains this

internal/karma/karma.conf.js
gregmagolan: Oh. I took a closer look at the MANIFEST file. I think this code can be improved to get rid the `build_bazel_rules_typescript/external/` bit. Will try it out.

gregmagolan: Cleaned this up. No `build_bazel_rules_typescript/external/` prefix is necessary.

internal/karma/karma.conf.js
gregmagolan: Cleaned up. No MANIFEST case is not needed.

WORKSPACE
alexeagle: nevermind, this is just a fast-forward of master

internal/tsc_wrapped/strict_deps.ts
alexeagle: yes, that's what I meant, that way we see the familiar `path.relative` below, and makes this delta smaller

gregmagolan: hmmm. this will conflict with the existing `path` imports and uses in `compiler_host.ts` and `tsconfig.ts`.

gregmagolan: I can just re-export the functions used out of './path'. That will be pretty clean.

internal/tsc_wrapped/path.ts
gregmagolan: Yup. I meant forward slashes. Updated to `preserves forward slashes on Windows`

achew22:
@gregmagolan, does this address #152?

gregmagolan:
> @gregmagolan, does this address #152?

Nope. Although there is a fix for that that @alexeagle has submitted internally

gregmagolan:
@achew22 Its been merged in. Fixed on rules_typescript master now. Commit is: Workaround for https://github.com/Microsoft/TypeScript/issues/22208

alexeagle:
Okay I got @mprobst on chat, here is the code review:

1) split into two PRs, one for tsc_wrapped and one for karma.
2) for tsc_wrapped, copy this function https://github.com/Microsoft/TypeScript/blob/a564912d9a335fe67ce7c5e4dabce46015bad8d4/src/compiler/core.ts#L1943 into compiler_host.ts
and then call it only in the minimal places required to make Windows green - things like the change in worker.ts are only error reporting so that can't be strictly required.

gregmagolan:
Replaced by #160, #161 and #162

PR #161, opened 2018-03-02
Windows fixes for examples tests

gregmagolan:
MANIFEST fixes for examples tests for Windows

/cc @alexeagle

### DIFF
--- a/examples/es5_output/es5_output_test.sh
+++ b/examples/es5_output/es5_output_test.sh
@@ -1,16 +1,39 @@
 #!/bin/bash
 set -e
 
+MANIFEST="$TEST_SRCDIR/MANIFEST"
+if [ -e "$MANIFEST" ]; then
+  while read line; do
+    declare -a PARTS=($line)
+    if [ "${PARTS[0]}" == "build_bazel_rules_typescript/examples/some_library/library.js" ]; then
+      readonly LIBRARY_JS=$(cat ${PARTS[1]})
+    elif [ "${PARTS[0]}" == "build_bazel_rules_typescript/examples/bar.js" ]; then
+      readonly BAR_JS=$(cat ${PARTS[1]})
+    elif [ "${PARTS[0]}" == "build_bazel_rules_typescript/examples/foo.js" ]; then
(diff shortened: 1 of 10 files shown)


examples/es5_output/es5_output_test.sh
```diff
@@ -1,40 +1,69 @@
...
 if [[ "$BAR_JS" != *"require(\"build_bazel_rules_typescript/examples/foo\")"* ]]; then
   echo "Expected bar.js to require named module foo, but was"
   echo "$BAR_JS"
   exit 1
 fi
 
 # should give a name to required modules from other compilation unit
-readonly FOO_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/bar.js)
-if [[ "$FOO_JS" != *"require(\"some-lib/library\")"* ]]; then
+if [[ "$BAR_JS" != *"require(\"some-lib/library\")"* ]]; then
```
alexeagle: why did this change from FOO to BAR? seems like a mistake?

gregmagolan: the original was wrong. code was actually pointing to bar.js: `readonly FOO_JS=$(cat $TEST_SRCDIR/build_bazel_rules_typescript/examples/bar.js)`

gregmagolan:
@mprobst Can you review when you have some time. This fixes the last CI failures on Windows.

gregmagolan:
@mprobst Please review. This fixes the last CI failures on Windows. Thanks

alexeagle:
I'll find some other googler to review. But looking at this again, it seems like we have way too much duplication of the logic to lookup a file possibly through the manifest. Any thoughts how we could make a shared .sh file that we source and call a function? Should be shared with e.g. https://github.com/bazelbuild/rules_nodejs/blob/master/internal/node/node_launcher.sh#L106 as well

gregmagolan:
https://github.com/bazelbuild/rules_nodejs/pull/207 adds a bash function that can be used to resolve runfiles. Will reduce the duplication here once that lands.

Issue #159, opened 2018-03-02
Failures on Bazel's CI


buchgr:
Starting with commit 65cc4da75bcd4cb88f7dcd621da17afc29c80ff3 `//examples/testing:testing` started to fail on macOS. It has been red ever since: https://buildkite.com/bazel/rules-typescript-typescript/builds/80#15f4a0cb-0aaa-446a-a017-3fe5ee70033b

Here's the error we get. Does this at all look related to the change @alexeagle?
```
exec ${PAGER:-/usr/bin/less} "$0" || exit 1
-----------------------------------------------------------------------------
[32m01 03 2018 15:46:00.651:INFO [karma]: [39mKarma v1.7.1 server started at http://0.0.0.0:9876/
[32m01 03 2018 15:46:00.654:INFO [launcher]: [39mLaunching browser ChromeHeadless with unlimited concurrency
[32m01 03 2018 15:46:00.663:INFO [launcher]: [39mStarting browser ChromeHeadless
[33m01 03 2018 15:47:00.662:WARN [launcher]: [39mChromeHeadless have not captured in 60000 ms, killing.
[33m01 03 2018 15:47:02.664:WARN [launcher]: [39mChromeHeadless was not killed in 2000 ms, sending SIGKILL.
[33m01 03 2018 15:47:04.666:WARN [launcher]: [39mChromeHeadless was not killed by SIGKILL in 2000 ms, continuing.
```

gregmagolan:
This is fixed with the karma fixes in PR #162. This PR switches karma to use the Chrome binary from puppeteer. @alexeagle is working on merging that in.

https://buildkite.com/bazel/rules-typescript-typescript/builds/97

buchgr:
Awesome! thanks @gregmagolan

buchgr:
Heya!

We are still seeing this error on macOS [1]. Any thoughts?

[1] https://storage.googleapis.com/bazel-buildkite-artifacts/d7a5dfab-0055-4534-9af4-c7981499ebf9/examples/testing/testing/attempt_1.log

alexeagle:
I don't see why that karma test is live on CI already.
https://github.com/bazelbuild/rules_typescript/commit/3bfca0624df383e31cff697c7a07d229f9179e05 only turned on `bazel build //examples:all`

lfpino:
Friendly ping @gregmagolan

gregmagolan:
[...]

PR #162, opened 2018-03-02
ts_web_test & karma fixes for Windows

gregmagolan:
ts_web_test and karma fixes for Windows

/cc @alexeagle


gregmagolan:
Also fixes #159

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,56 @@
...
+// environment variable. This line points Karma to use puppeteer instead.
+// See
+// https://github.com/karma-runner/karma-chrome-launcher/blob/master/README.md#headless-chromium-with-puppeteer
+process.env.CHROME_BIN = require('puppeteer').executablePath();
+
+let files = [
+  TMPL_files
+];
+
+const manifestFile = path.join(process.env.TEST_SRCDIR, 'MANIFEST');
```
alexeagle: could you add docs on what the code below does?

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,56 @@
...
+if (fs.existsSync(manifestFile)) {
```
alexeagle: if for some misconfiguration this file is missing, will we just silently fail in an odd way? would it be better to rather fail early?

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,56 @@
...
+  const manifest = {};
```
alexeagle: what's in this struct? what's in the manifest? docs?

alexeagle: what's in this struct? what's in the manifest? docs?

internal/karma/karma.conf.js
```diff
@@ -1,5 +1,56 @@
...
+  for (l of fs.readFileSync(manifestFile, 'utf8').split('\n')) {
+    const m = l.split(' ');
+    manifest[m[0]] = m[1];
+  }
+  const manifestKeys = Object.keys(manifest);
+  files = files.map((f) => {
+    if (manifestKeys.includes(f)) {
```
alexeagle: this is O(n^2), but should be easy to avoid using a map

internal/karma/karma.conf.js
gregmagolan: The file is expected to be missing in linux and OSX. It should only be present on Windows.

internal/karma/karma.conf.js
gregmagolan: I'll add some comments. I couldn't find anything on the MANIFEST file in the Bazel docs.

internal/karma/karma.conf.js
gregmagolan: 👍

internal/karma/karma.conf.js
gregmagolan: Thanks. Updated to remove the double lookup. It looks like O(n*m) to me where n is the number of test files and m is the number of files in the MANIFEST.

gregmagolan:
@mprobst Please review when you have some time

PR #163, opened 2018-03-03
Example pulls nested file in genrule.

achew22:
Rules can make subdirs, add a demo of pulling a var from a rule that
makes subdirs.

I wrote this to validate a theory on a bug I was having. I was expecting
this to fail, but it didn't so I figured I'd just commit it and improve the
examples a little.


alexeagle:
sorry, ran out of energy to get this re-based, I don't think it's that important as Bazel users will have to know about subdirs

PR #184, opened 2018-04-04
feat(tsetse): add `ban_nested_it` rules

alan-agius4:
Bans nested 'it' methods,  as these cause falsy positive in unit tests.

To test It I modified the BUILD.bazel to 
```ts
ts_library(
    name = "test_lib",
    srcs = glob(["tsetse/tests/ban_nested_it/*.ts"]),
    tsconfig = "//internal:tsc_wrapped/tsconfig.json",
)
```

As it seems that `tsetse/tests` are being ignored from `bazel test`

This is following a convo with @alexeagle  in https://github.com/angular/angular/pull/23153, couldn't help it in trying to create a rule :)


alan-agius4:
@alexeagle, I have added the documentation for rule.

I have checked and at the moment only `errors` are supported: https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsetse/failure.ts#L21

Based on this doc http://errorprone.info/docs/criteria, It seems that this rule should be more of an `Error` since it not a style guide violations, potential serious bugs, or performance issue. But a actual issue since it causes falsy positives, and should be deemed mandatory to fix.

I can implement that `Warnings` diagnostics are supported of course. Either in this PR if you deem this needs to be a warning or in another PR if you think it might be used in the future.

docs/ban_nested_it.md
```diff
@@ -0,0 +1,44 @@
+## Expect not to have nested 'it' call expressions
```
alexeagle: somewhere it needs to say what `it` we are talking about - is it jasmine?

internal/tsetse/rules/ban_nested_it.ts
```diff
@@ -0,0 +1,57 @@
...
+	readonly code = ErrorCode.BAN_NESTED_IT;
+
+	register(checker: Checker) {
+		checker.on(ts.SyntaxKind.CallExpression, checkCallExpression, this.code);
+	}
+}
+
+function checkCallExpression(checker: Checker, node: ts.CallExpression) {
+	const identifier = (node.expression as ts.Identifier);
+	if (identifier.getText() === 'it' && hasNestedItExpression(node)) {
```
alexeagle: I think we should resolve the symbol to which module defines it - otherwise someone writing their own method called `it` would have this check trigger

docs/ban_nested_it.md
```diff
@@ -0,0 +1,44 @@
...
+
+Nested `it` call expression in spec files will cause false positives for unit tests.
+
+To fix this, the the upper `it` call expression should be changed to a `describe`.
+
+### Postive Examples:
```
alexeagle: positive misspelled

Also it would be awesome if we used the executable test cases to generate this section of the docs (I did that for Error Prone) - for later :)

internal/tsetse/rules/ban_nested_it.ts
alan-agius4: Yeah good point. Give me some time and I'll do it.

docs/ban_nested_it.md
alan-agius4: fix the spelling

internal/tsetse/rules/ban_nested_it.ts
alan-agius4: Do you know any other way to get the `moduleName` other than

```ts
const signature = checker.typeChecker.getResolvedSignature(node);
const file = signature.getDeclaration().getSourceFile().moduleName;
```

As the moduleName here it always being returned `undefined`, but If I try to use `.fileName` instead I get the correct file path. 

Cause I was not going to do something like;

```ts
const signature = checker.typeChecker.getResolvedSignature(node);
const file = signature.getDeclaration().getSourceFile().fileName;

if (file.indexOf('jasmine') >= 0  && node.getText() === 'it') 
```

I don't like the indexOf implementation that much, but cannot find a way to get the moduleName since it's being returned undefined.

alan-agius4: This has been addressed as well

bowenni:
That sounds like a bug fixed in Jasmine 2.9
Issue: https://github.com/jasmine/jasmine/issues/1295
Fixed: https://github.com/jasmine/jasmine/pull/1411

Maybe upgrade jasmine in angular/angular (currently at 2.4)?
Also I'd like to avoid framework specific checks in tsc.

alan-agius4:
I can try to update angular/angular

PR #185, opened 2018-04-06
Handle module names that come from vanilla ngc compiler ngsummary.json files

gregmagolan:


internal/common/module_mappings.bzl
```diff
@@ -70,11 +70,10 @@ def get_module_mappings(label, attrs, srcs = [], workspace_name = None, mappings
     mn = attrs.module_name
     if not mn:
       mn = label.name
-    mr = label.package
-    if workspace_name:
-      mr = "%s/%s" % (workspace_name, mr)
-    elif label.workspace_root:
-      mr = "%s/%s" % (label.workspace_root, mr)
+    mr = "/".join([p for p in [
```
gregmagolan: This is just a little cleanup. If `label.package` is "" then `mr` ends up with a double forward-slash in it

gregmagolan:
After discussing with @alexeagle, we'll handle this issue elsewhere

PR #186, opened 2018-04-06
Prevent double forward slash from creeping into module_mappings when label.package is empty

gregmagolan:


gregmagolan:
@mprobst @alexeagle Please review. Thanks.

PR #189, opened 2018-04-12
Exclude bootstrap and node_modules files from possible karma test files

gregmagolan:
@alexeagle This fixes ts_web_test failures in angular-bazel-example since some files from deps in node_modules that end in `test.js` are mistaken as test files to run.

Thinking about this, it seems fragile to look through all the `deps` of ts_web_test to find files that end in `test.js` and `spec.js` (what the original karma `requirejs.config.tpl.js` does). I think it would be more robust if ts_web_test had a separate attribute besides deps to specify the test files themselves. Maybe `test_deps` or `test_files`? Thoughts?

### DIFF
--- a/internal/karma/karma.conf.js
+++ b/internal/karma/karma.conf.js
@@ -11,7 +11,8 @@ const tmp = require('tmp');
 process.env.CHROME_BIN = require('puppeteer').executablePath();
 
 let files = [
-TMPL_files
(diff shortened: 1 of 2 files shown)


gregmagolan:
On further investigation. This seems like a good solution. User test files should come from node_modules. I also excluded the bootstrap scripts from the list of possible karma test files.

PR #193, opened 2018-04-27
Add ts_proto_library rule.

### LINKED ISSUE #61: Demonstrate usage of protocol buffers
Should have type-safe access to generated protos.

We could use this in our own repo: currently proto objects are type `any`, eg
https://github.com/bazelbuild/rules_typescript/blob/64567904e3648b728f356057bdf6d0ed603be388/internal/tsc_wrapped/worker.ts#L26

https://github.com/dcodeIO/ProtoBuf.js/#usage-with-typescript seems like the obvious `.d.ts` code generator to use. This also implies that the library must use the protobuf.js generated runtime.

/cc @jart

alexeagle:
Add example that uses it under a karma test.

Fixes #61 

/cc @kellycampbell


gregmagolan:
Fix windows CI?

examples/protocol_buffers/car.proto
```diff
@@ -0,0 +1,11 @@
+syntax = "proto3";
+
+import "examples/protocol_buffers/tire.proto";
+
+message Car {
+  string make = 1;
+  string model = 2;
+  int32 year = 3;
```
mrmeku: You might consider exercising your test with a message containing an int64. `pbjs` has a couple of modes for handing `int64` in JS and one of the requires that the `long` node_module be used. We should test that we integrated that scenario correctly

examples/tsconfig.json
```diff
@@ -1,7 +1,8 @@
 {
   "compilerOptions": {
     "strict": true,
-    "lib": ["es2015.promise", "dom", "es5"]
+    "lib": ["es2015.promise", "dom", "es5"],
+    "rootDirs": [".", "../bazel-bin/examples"]
```
mrmeku: Is this for editor support of generated files?

cgrushko:
cc @iirina

examples/protocol_buffers/car.proto
alexeagle: thanks, good point

internal/devserver/ts_devserver.bzl
```diff
@@ -106,6 +116,7 @@ ts_devserver = rule(
         "serving_path": attr.string(),
         "data": attr.label_list(allow_files = True, cfg = "data"),
         "static_files": attr.label_list(allow_files = True),
+        "bootstrap": attr.label_list(allow_files = [".js"]),
         # User scripts for the devserver to concat before the source files
         "scripts": attr.label_list(allow_files = True),
```
gregmagolan: Suppose it is correct to limit scripts to [".js"] as well

internal/protobufjs/ts_proto_library.bzl
```diff
@@ -0,0 +1,150 @@
...
+  # Return a structure that is compatible with the deps[] of a ts_library.
+  return struct(
+    files = depset([dts]),
+    typescript = struct(
+      declarations = [dts],
+      transitive_declarations = [dts],
+      type_blacklisted_declarations = [],
+      es5_sources = depset([js_es5]),
+      es6_sources = depset([js_es6]),
+      transitive_es5_sources = depset(),
```
jbedard: I tried this PR out and had issues because the transitive sources here don't include the normal sources. I think transitive should be direct + indirect sources? In my case this causes issues when this read by [collect_es6_sources](https://github.com/bazelbuild/rules_nodejs/blob/befbd8961c18de98e254bea94db08ede78889f86/internal/common/collect_es6_sources.bzl#L36) which only reads `transitive_es6_sources`...

alexeagle: Thanks! I didn't test it in a production bundle, that's missing here...

alexeagle:
Carmi, what would such an aspect do? Is there actually an action to cache
at each proto_library node in the graph? In this case, the proto compiler
does a global compilation over all the transitive sources so I don't think
it would be an improvement, and adds complexity.

cgrushko:
For all languages, protoc reads all transitive sources, and this step is
unavoidable (due to protoc architecture) but also super cheap.

OTOH, compiling the TS files themselves could he expensive. Java and C++
are expensive enough to be worth it. Remember that the number of compiled
files is quadratic in the depth of the dependency tree.

The bigger problem is ODR, though - without aspects you will end up with
the same symbols generated by multiple rules. Depending on the language
this could lead to linking errors (e.g. C++) or output file conflicts if
the name of the output is derived from the names of the .proto files.

Aspects also allow parallelism when compiling the generated files, vs
transitive compilation.

PR #198, opened 2018-05-07
Remove puppeteer & turn on preserve-symlinks

gregmagolan:
Removed dependency on puppeteer which was breaking preserve-symlinks since Chrome was downloaded into node_modules with invalid file names.

Platform dependent chromium archive is now downloaded via repository rule. ts_webtest unzips this file and uses this chrome to run the webtest.

--preserve-symlinks is also turned on for the repository.


gregmagolan:
Replaced by #199

PR #204, opened 2018-05-15
Update to latest rules_nodejs & cleanup

gregmagolan:


.circleci/bazel.rc
```diff
@@ -0,0 +1,28 @@
+# These options are enabled when running on CI
```
alexeagle: can you also remove the `--config=ci` setup? Both anywhere that is passed and also any config options in bazel.rc?
This setup should be mutually exclusive with using `--config`

.circleci/config.yml
```diff
@@ -9,13 +9,19 @@
...
+var_1: &docker_image angular/ngcontainer:0.3.0
+var_2: &cache_key rules_typescript-{{ checksum "yarn.lock" }}-0.3.0
 var_3: &setup-bazel-remote-cache
   run:
     name: Start up bazel remote cache proxy
     command: ~/bazel-remote-proxy -backend circleci://
     background: true
+# Move node binaries out of the way to enforce that Bazel uses
+# only the hermetic ones it downloads
+var_4: &hide_node_and_yarn_local_binaries
```
alexeagle: there's a nicer pipe syntax; switch this to

```
var_4: &hide_node_and_yarn_local_binaries
  run:
    name: Move node, npm, and yarn binaries
    command: |
      sudo mv /usr/local/bin/node /usr/local/bin/node_
      sudo mv /usr/local/bin/npm /usr/local/bin/npm_
      sudo mv /usr/local/bin/yarn /usr/local/bin/yarn_
```

examples/protocol_buffers/BUILD.bazel
```diff
@@ -55,10 +55,48 @@ ts_devserver(
...
 
 # Test for production mode
-load("@build_bazel_rules_nodejs//:defs.bzl", "rollup_bundle")
+load("@build_bazel_rules_nodejs//:defs.bzl", "rollup_bundle", "nodejs_binary")
 
 rollup_bundle(
     name = "bundle",
     deps = [":app"],
     entry_point = "examples/protocol_buffers/app",
+    globals = {
```
alexeagle: this map is already defined in the `js_library` - any idea how we could re-use it?

examples/protocol_buffers/BUILD.bazel
```diff
@@ -55,10 +55,48 @@ ts_devserver(
...
+        "long": "Long",
+        "protobufjs/minimal": "protobuf",
+    },
+)
+
+# Needed because the prodserver only loads static files that appear under this
+# package.
+genrule(
+    name = "protobufjs",
+    srcs = ["@build_bazel_rules_typescript_protobufs_compiletime_deps//:node_modules/protobufjs/dist/minimal/protobuf.min.js"],
```
alexeagle: isn't there a filegroup that gives both files? can we have one genrule that uses that?

package.json
```diff
@@ -20,15 +20,20 @@
...
         "protractor": "^5.2.0",
         "tsickle": "0.25.x",
         "tsutils": "2.20.0",
         "typescript": "2.7.x"
     },
     "scripts": {
-        "pretest": "webdriver-manager update && bazel build examples/app:all",
-        "test": "concurrently \"bazel run examples/app:devserver\" protractor --kill-others --success first",
+        "pretest": "webdriver-manager update $CHROMEDRIVER_VERSION_ARG && bazel build --config=ci examples/app:e2e && bazel build --config=ci examples/protocol_buffers:e2e",
+        "test": "yarn test-examples-app-devserver && yarn test-examples-app-prodserver && yarn test-examples-protobuf-devserver && yarn test-examples-protobuf-prodserver",
```
alexeagle: this is kinda gross - can't we keep the protractor config in the root and tell it about multiple tests to run?

.circleci/bazel.rc
gregmagolan: You mean the `build --local_resources=14336,8.0,1.0`?

.circleci/config.yml
gregmagolan: Thanks

package.json
gregmagolan: Good idea. I made it work with a single protractor config making use of `suites`. The serving port has to be the same so I added a `port` attribute to ts_devserver. Can't think of how to remove the multiple package.json scripts unless test is just changed to a shell script that run the tests serially.

.circleci/config.yml
```diff
@@ -90,6 +105,7 @@ jobs:
       - checkout:
           <<: *post_checkout
       - run: .circleci/setup_cache.sh
+      - run: sudo cp .circleci/bazel.rc /etc/bazel.bazelrc
       - *setup-bazel-remote-cache
       # To get the skylint binary from the cache, we need to build with --config=ci
```
alexeagle: you can remove next 5 lines

.circleci/bazel.rc
alexeagle: yeah all of them, we should remove the `build:ci` lines from `tools/bazel.rc` since we now have a separate file for these

examples/protocol_buffers/BUILD.bazel
alexeagle: okay let's not wait too long :)

examples/protocol_buffers/BUILD.bazel
alexeagle: there must be a way! but it's okay if not

package.json
alexeagle: sure

.circleci/bazel.rc
gregmagolan: Done 👍

examples/protocol_buffers/BUILD.bazel
gregmagolan: Found a way

package.json
gregmagolan: On second thought. If changed to a shell script then the e2e tests won't run on windows :(

examples/protocol_buffers/BUILD.bazel
```diff
@@ -52,13 +52,50 @@ ts_devserver(
...
     name = "bundle",
     deps = [":app"],
     entry_point = "examples/protocol_buffers/app",
+    globals = {
+        "long": "Long",
+        "protobufjs/minimal": "protobuf",
+    },
+)
+
+# Needed because the prodserver only loads static files that appear under this
```
alexeagle: shouldn't be needed anymore right? since we have `additional_root_paths`?

Issue #131, opened 2018-01-22
ts_web_test should serve static files


alexeagle:
Some tests need karma to be configured such that static files (e.g some JSON data) can be served with a window.fetch.
For example, Angular has some disabled tests (`tags = ["manual"]`) because this isn't available.

The obvious design is to be similar to `ts_devserver` to keep a simple mental model for users. Accept the `data[]` attribute and serve anything in data, relative to the package where the rule appears.

alexeagle:
@jelbourn would like this as well, so that angular components with `templateUrl` could use JIT compiler under ts_web_test

gregmagolan:
I think it should also allow for additional_root_paths as ts_devserver handles in #177

gregmagolan:
Resolved with #209 (https://github.com/bazelbuild/rules_typescript/commit/80ce606dcb0076378629cc34d96852302d17f8ce)

PR #209, opened 2018-05-24
Add static_files attribute to ts_web_test & ts_web_test_suite

gregmagolan:
Allows re-enabling `packages/upgrade/test:test_web` test in angular/angular: https://github.com/gregmagolan/angular/tree/ts-web-test-static-files (based off of https://github.com/angular/angular/pull/23845)


gregmagolan:
Resolves #131

PR #214, opened 2018-06-07
Replace deprecated `library` with `embed` in go_test

alex-vlasov:
`library` attribute is deprecated in `go_test`:
```
DEPRECATED: //internal/devserver:devserver_test : the library attribute on go_test is deprecated. Please migrate to embed.
```
In rules_go v.0.12.0 it's causing an error:
```
(18:04:54) ERROR: /home/alex/.cache/bazel/_bazel_alex/9def434882dfecce7f21cdef6844c099/external/build_bazel_rules_typescript/internal/concatjs/BUILD:18:1: @build_bazel_rules_typescript//internal/concatjs:concatjs_test: no such attribute 'library' in 'go_test' rule
```

### DIFF
--- a/internal/concatjs/BUILD
+++ b/internal/concatjs/BUILD
@@ -21,5 +21,5 @@ go_test(
     importpath = "do_not_import",
     size = "small",
     srcs = ["concatjs_test.go"],
-    library = ":concatjs",
+    embed = [":concatjs"],
 )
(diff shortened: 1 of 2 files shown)


alexeagle:
Hmm, this breaks us under Blaze internally at Google, I'm checking with @jayconrod to see what to do.

alexeagle:
Sorry this is stuck, the go rules are not the same internally as under Bazel so it requires some additional mechanics to make the code portable :(
Will get to it next week I hope

Issue #216, opened 2018-06-08
nodejs_binary depending on ts_library should get higher-level syntax


Toxicable:
Devs should have a way to configure ts_library 's JS target.

When `ts_library` is being used for Nodejs applications the target is a bit different than browser applications.
For a browser app you always want to use es5 since you don't know what browser your app is going to be run on. 
But with Nodejs you can pick the Nodejs version and therefore know ahead of time what target it supports.
Therefore by picking a higher target you can often gain performance from your application, depending on use case.

Another reason for allowing variable targets is interpoerablity with other nodejs modules. One such example of not supporting es5 is the popular ORM Sequelize https://github.com/RobinBuschmann/sequelize-typescript/blob/master/lib/models/BaseSequelize.ts#L14

alexeagle:
Currently we have exactly two output flavors from TypeScript rules, and @mprobst has pushed back when we suggested adding more (Angular has a need for "ESM5" which we have a hacky way to produce in 
https://github.com/angular/angular/blob/master/packages/bazel/src/esm5.bzl

I'd rather avoid the complexity of having more output flavors. If we can keep nodejs programs using the "devmode" (non-optimizer/bundler friendly) output, this keeps our system simpler.

The simple resolution to this issue seems to be raising the `--target` of devmode compilations to `es2015`, we just need to confirm we don't break users who rely on ES5 syntax in the browser for whatever reason.

(Note, the devmode output is not intended to ship to end-user browser runtime, so there's no constraint there)

alexeagle:
es2015 devmode is now the default, it applies to nodejs inputs too

thesayyn:
It is awkward that why I can't use a different module and target value in my tsconfig since I'm not depending on ts_devserver. the current options prevent me from enabling resolveJsonModule because it requires a commonjs & es2015 target.  Maybe we can add an option to let the advanced user fly freely.

marcus-sa:
^ I agree.
I've gone back and thought about moving completely over to Bazel, since you can't even define the target for TypeScript...

Toxicable:
You can specify an override the default target with `devmodeTargetOverride` in your tsconfig
```
  "bazelOptions": {
    "devmodeTargetOverride": "esnext"
  },
```
Also note that at runtime we use the `devmode` artifacts, which default to es2015 for nodejs, `prodmode` artifacts are used for web bundling targets.

alexeagle:
[...]

PR #217, opened 2018-06-09
feat(ts_library): bump default target to es2015

Toxicable:
closes https://github.com/bazelbuild/rules_typescript/issues/216

If this is accepted a side effect is that all instances `for .. of` will no longer be downleveled into `for` loops where they're compiled to es2015.
The downside of this is that `for .. of` is significantly slower than `for` loops.
So to no rely on downleveling we should replace all instances of for of with plain for loops to ensure we're always getting the most performance out of them.
I've made the same change in Angular here: https://github.com/angular/angular/pull/24534


internal/common/tsconfig.bzl
```diff
@@ -161,7 +161,7 @@ def create_tsconfig(ctx, files, srcs,
   # Keep these options in sync with those in playground/playground.ts.
   compiler_options = {
       # De-sugar to this language level
-      "target": "es5" if devmode_manifest or ctx.attr.runtime == "nodejs" else "es6",
+      "target": "es2015",
```
mrmeku: I think you need to add back in the ternary so that es6 sources can still be generated. As I understand things es2015 is a strict subset of es6. If I'm wrong about that feel free to tell me :p

internal/tsc_wrapped/tsconfig.json
```diff
@@ -10,8 +10,9 @@
       "es5",
       "es2015.core",
       "es2015.collection",
+      "es2015.promise",
+      // TODO: remove "es2015.iterable" this and replace all instances of for .. of with for
```
mrmeku: Here is a formatted version of this comment.
// TODO: Remove "es2015.iterable" by replacing all instances of "for .. of".

I think you should also track this TODO within a github issue where it is explained more throughouly. As things read right now, I'm not sure who replacing for of loops will let you replace that particular lib

internal/tsetse/tsconfig.json
```diff
@@ -9,7 +9,10 @@
       "es5",
       "es2015.core",
       "es2015.collection",
-      "es2015.promise"
+      "es2015.promise",
+      "es2015.symbol",
+      // TODO: remove this and replace all instances of for .. of with for
```
mrmeku: See my previous comment

internal/tsc_wrapped/tsconfig.json
Toxicable: Added an explainer in the PR description. 
Ill fix up this issue if we go through with this PR.

internal/common/tsconfig.bzl
Toxicable: Yup you were totally right.
Also In putting that back we're also able to maintain backwards compatibility aswell since non of the tests have to be edited

examples/dev_output/dev_output_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+set -e
+
+# should produce named UMD modules
```
alexeagle: This comment looks out-of-date

internal/common/tsconfig.bzl
```diff
@@ -161,7 +161,7 @@ def create_tsconfig(ctx, files, srcs,
   # Keep these options in sync with those in playground/playground.ts.
   compiler_options = {
       # De-sugar to this language level
-      "target": "es5" if devmode_manifest or ctx.attr.runtime == "nodejs" else "es6",
+      "target": "es2015" if devmode_manifest or ctx.attr.runtime == "nodejs" else "es6",
```
alexeagle: `es2015` is just an alias for `es6`

https://github.com/Microsoft/TypeScript/blob/7a322d72f76f10e4433905853f6fe1370d98ea83/src/parser/commandLineParser.ts#L150

you say there was some behavior difference when you restored the trinary? I don't see how that's possible...

internal/tsc_wrapped/tsconfig.json
```diff
@@ -10,8 +10,9 @@
       "es5",
       "es2015.core",
       "es2015.collection",
+      "es2015.promise",
+      // TODO: Remove "es2015.iterable" by replacing all instances of "for .. of".
```
alexeagle: This needs some discussion. I don't see how we would practically change user behavior.

I don't think it's our job to tell all users how to write ES2015 that performs well on all platforms. They'd have this problem even without using any Bazel stuff.

Presumably the browsers get faster at running `for..of` loops in the future, then users change back? Seems like churn.

Toxicable: This is mainly for the performance of tsc_wrapped, in the vast majority of use cases the you'll never see a practical change in execution time, unless it's something like a compiler wrapper.

Not to mention you'll gain much more perf in all the other areas.

For .. of will never be faster the for.

examples/dev_output/dev_output_test.sh
Toxicable: whops, copy paste error, fixed

internal/common/tsconfig.bzl
Toxicable: I just tried 3 tests doing `bazel test ...`: `"target": "es6",`, `"target": "es2015",`, `"target": "es2015" if devmode_manifest or ctx.attr.runtime == "nodejs" else "es6",`
es6 fails with
>Expected library.js to contain 'export const cool = 1;' but was

es2015 fails with
>external/build_bazel_rules_typescript/internal/tsetse/language_service_plugin.ts(37,1): error TS1203: Export assignment cannot be used when targeting ECMAScript modules. Consider
using 'export default' or another module format instead.

So only the turnery is passing all tests.
My guess is that it looks like downstream rules are basing their `"module"` off a string comparison of the `target`, comp[lete guess but obth of those errors looks like `module` issues (edited)

Toxicable:
@alexeagle the es5_output test has no test to check for downleveled classes.

Toxicable:
I've added in the for of -> for loop conversions and done in the same in Angular
https://github.com/angular/angular/pull/24534

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -199,7 +201,9 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
    * @return the path without any rootDirs, eg. path/to/file
    */
   private rootDirsRelative(fileName: string): string {
-    for (const root of this.options.rootDirs) {
+    const rootDirs = this.options.rootDirs;
```
alexeagle: nit: why bother declaring a local variable here?

internal/common/tsconfig.bzl
```diff
@@ -161,7 +161,7 @@ def create_tsconfig(ctx, files, srcs,
   # Keep these options in sync with those in playground/playground.ts.
   compiler_options = {
       # De-sugar to this language level
-      "target": "es5" if devmode_manifest or ctx.attr.runtime == "nodejs" else "es2015",
+      "target": "es2015",
```
alexeagle: note: this is shared with google3 which we aren't ready to flip yet, so I think this should go in `internal/build_defs.bzl` where we override the results from `tsc_wrapped_tsconfig`

internal/tsc_wrapped/compiler_host.ts
Toxicable: Just being explicit that there's no function changes here by using the variable in the same way.
Happy to refactor if you like

examples/dev_output/dev_output_test.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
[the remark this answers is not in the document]
Toxicable: done

internal/common/tsconfig.bzl
Toxicable: done, `internal/common/tsconfig.bzl` is now unchanged

Toxicable:
Im not sure how this failure relates to my changes:
>yarn jasmine internal/e2e/default_tsconfig_test.js
>ERROR: /.../external/build_bazel_rules_typescript/internal/BUILD.bazel:31:1: every rule of type nodejs_binary implicitl
y depends upon the target '//:node_modules', but this target could not be found because of: no such target '//:node_modules': target 'node_modules' not declared in package '' defined by /...

internal/tsc_wrapped/compiler_host.ts
Toxicable: Oh wait, my bad, that statement applies to all other for loop replacements.
This instance was to avoid going through `this` and `options` on each access, which can be slow depending on if it's a hot path and if `options` is mono or mega morphically tracked. So this is the safe option
Some notes here on the topic: https://github.com/angular/angular/blob/master/packages/core/src/render3/PERF_NOTES.md#monomorphic-vs-megamorphic-code

alexeagle:
@Toxicable that was a build breakage on master, if you just rebase your PR should go green.

@mprobst please review here and approve, then I'll make you the reviewer of the sync CL.

Toxicable:
@mprobst WHats the plan here, are we still trying to get this in?

internal/tsc_wrapped/file_cache.ts
```diff
@@ -103,12 +103,16 @@ export class FileCache<T = {}> {
     this.debug('updating digests:', digests);
     this.lastDigests = digests;
     this.cannotEvict = false;
-    for (const [filePath, newDigest] of digests.entries()) {
+    const fileEntries = Array.from(digests.keys());
+    for (let i = 0; i < fileEntries.length; i++) {
```
IgorMinar: have we measured that `for of` is actually still slow in v8? 

v8 team has been focused on perf improvements of common js patterns:
- https://v8.dev/blog/fast-for-in
- https://v8.dev/blog/spread-elements

have we measured this using the perf harness that @filipesilva's built?

Toxicable: Interesting, looks like im pretty out of date on this one.
Looking at this here: https://jsperf.com/for-of-vs-for-loop - specifically the section at the bottom showing this test on Chrome versions going back to 49 when `for` was back then significantly faster, you can see the improving trend over the past few versions which would aling with those articles.

Right now on Chrome 70 on my system, it appears that `for of` is actually ~20% faster.
Which I wouldn't take as a hard fact that `for of` is 20% faster than `for`, just that my original statement of `for` being significantly faster is incorrect.

IgorMinar: I think we can safely conclude that we shouldn't be avoiding for-of because of perf concerns. Can you please update this PR? thanks!

PR #222, opened 2018-06-18
Update to typescript 2.8.4 and tsutils 2.27.1

gregmagolan:


gregmagolan:
@vikerman This is to update the rules_typescript typescript version to match angular. I ran into a build error with the version skew in https://github.com/angular/angular/pull/24370 in the intergration test

alexeagle:
Hmm, then the version is different from what we use to build tsc_wrapped in google3. I'm not sure which one we should aim to match. Any other way to fix that version skew?

gregmagolan:
Figured out the root cause, which was that the bazel integration example had typescript 2.8.4 in its package.json but not the matching tsutils (so the tsutils from rules_typescript was being used). I had suspected that and tried to add it earlier and still found the failure but I hadn't done a `bazel clean --expunge` so something wasn't rebuilt that needed to have been.

This rules_typescript upgrade no longer needed.

achew22:
@gregmagolan does `bazel clean --expunge` needed to be run very often for you? Is there a bug I can track to follow progress toward an `--expunge` free world?

gregmagolan:
@achew22 So `bazel clean` is sufficient in this case. Maybe what is happening is that the node_modules filegroup is not picking up the added `tsutils` folder after running yarn since its only watching the existing files.

```
filegroup(
    name = "node_modules",
    srcs = glob(["node_modules/**/*"])
)
```

achew22:
That's reassuring. Thanks for following up with that. `--expunge` is a painful thing to have to do in my repo

gregmagolan:
@achew22 Its saves a lot of time if you turn on the experimental_repository_cache in your .bazelrc: `build --experimental_repository_cache=/Users/greg/bazel_cache`

PR #223, opened 2018-06-25
Support for proper umd module naming for building external repos from source (such as angular/angular)

gregmagolan:


internal/tsc_wrapped/compiler_host.ts
```diff
@@ -332,6 +333,28 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
       }
     }
 
+    // TODO(gmagolan): clean this up
```
gregmagolan: This is basically doing a reverse lookup from the tsconfig paths to figure out what the module name should be. When building angular directly from source not all imports are scoped to the module name. The .ngsummary and .ngfactory files from the angular packages are not.

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -310,6 +310,7 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
     // named by code in that repository.
```
alexeagle: this function is getting quite complex, could you add unit tests for it pls?

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -335,6 +336,45 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
       }
     }
 
+    // The following handles the case when the filename is outside of
+    // bazelOpts.package but still needs to be mapped to the correct
```
alexeagle: it seems like the logic above that uses `bazelOpts.package` was too specific. Can this general logic replace some of that?

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -335,6 +336,45 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
...
+    // AMD module name.
+    //
+    // For example, when building Angular from source in downstream projects
+    // the fileName may be 'packages/common/src/pipes/date_pipe.ngfactory'
+    // while the bazelOpts.moduleName is '@angular/platform-browser' and the
+    // bazelOpts.package is 'packages/platform-browser'. We still want to map
```
alexeagle: Here's the fundamental problem:
if the file comes from another package, it should have been compiled already in that package
that means we should be seeing `.d.ts` files for the angular code that we are trying to name
`.d.ts` files should always have a `///<amd module-name="the correct name"/>` pragma at the top
so I don't see why we should ever be in the business of creating a module name for code that came from another package.

Does that make sense? what part of that is going wrong?

gregmagolan: Looking into it. It only happens for ngfactory files from angular. Here's the callstack when `CompilerHost.amdModuleName` is called for the ngfactory file `external/angular/packages/common/src/location/platform_location.ngfactory` when compiling `platform-browser` (bazelOpts.packageName is `packages/platform-browser`):

```
Mapped external/angular/packages/common/src/location/platform_location.ngfactory => @angular/common/src/location/platform_location.ngfactory (package packages/platform-browser) Error
    at /private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsc_wrapped/compiler_host.js:366:144
    at Array.forEach (<anonymous>)
    at _loop_1 (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsc_wrapped/compiler_host.js:361:47)
    at CompilerHost.amdModuleName (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/build_bazel_rules_typescript/internal/tsc_wrapped/compiler_host.js:376:25)
    at CompilerHost.ngHost.fileNameToModuleName (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/bazel/src/ngc-wrapped/index.js:255:31)
    at AotCompiler._fileNameToModuleName (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:546:28)
    at Object.importExpr (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:524:42)
    at /private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:439:90
    at Array.forEach (<anonymous>)
    at AotCompiler._createSummary (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:438:22)
    at AotCompiler._compileImplFile (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:382:77)
    at /private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:374:74
    at Array.map (<anonymous>)
    at AotCompiler.emitAllImpls (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler/src/aot/compiler.js:374:39)
    at AngularCompilerProgram.generateFilesForEmit (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler-cli/src/transformers/program.js:736:46)
    at AngularCompilerProgram._emitRender2 (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler-cli/src/transformers/program.js:330:27)
    at AngularCompilerProgram.emit (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler-cli/src/transformers/program.js:236:22)
    at Object.performCompilation (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/compiler-cli/src/perform_compile.js:194:29)
    at compile (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/bazel/src/ngc-wrapped/index.js:294:21)
    at runOneBuild (/private/var/tmp/_bazel_greg/2c2a834fcea131eff2d962ffe20e1c87/execroot/angular_bazel_example/bazel-out/host/bin/external/angular/packages/bazel/src/ngc-wrapped/index.js:94:27)
```

gregmagolan:
Replaced by principled for https://github.com/angular/angular/pull/25335

gregmagolan:
https://github.com/angular/angular/pull/25335 has unexpected side-effect breaking the angular-cli.

Looking at this PR again. I dug through the AOT compiler to see exactly where these filenames that are not able to be mapped to module names are coming from.

This happens during the generation step for `.ngfactory` files during an `ng_module` compile.

For example, when generating `/execroot/bazel_integration_test/external/angular/packages/common/http/src/interceptor.ngfactory.ts`, the compiler determines that it needs to import `InjectionToken` from `execroot/bazel_integration_test/bazel-out/darwin-fastbuild/bin/external/angular/packages/common/src/dom_tokens.ngfactory.d.ts`. So it tries to convert `_fileNameToModuleName`.

in `compiler.ts`:
```
private _fileNameToModuleName(importedFilePath: string, containingFilePath: string): string {
    return this._summaryResolver.getKnownModuleName(importedFilePath) ||
        this._symbolResolver.getKnownModuleName(importedFilePath) ||
        this._host.fileNameToModuleName(importedFilePath, containingFilePath);
  }
```

`this._summaryResolver.getKnownModuleName` and `this._symbolResolver.getKnownModuleName` both don't know what the module name of `angular/packages/common/src/dom_tokens.ngfactory.d.ts` is. When compiling `@angular/common`, these symbols may have been in the compiler's cache but at this point we are compiling `@angular/common/http` and the AOT compiler is generating an `.ngfactory` file in `@angular/common/http` that it knows needs an import from `@angular/common` but at this point it only has the filePath of the import and not the import module name.

So the AOT compiler asks the typescript compiler host to convert the filename to a module name. This ends up the bazel ngc-wrapped code here:

```
  ngHost.fileNameToModuleName = (importedFilePath: string, containingFilePath: string) => {
    if ((compilerOpts.module === ts.ModuleKind.UMD || compilerOpts.module === ts.ModuleKind.AMD) &&
        ngHost.amdModuleName) {
      return ngHost.amdModuleName({ fileName: importedFilePath } as ts.SourceFile);
    }
    const result = relativeToRootDirs(importedFilePath, compilerOpts.rootDirs).replace(EXT, '');
    if (result.startsWith(NODE_MODULES)) {
      return result.substr(NODE_MODULES.length);
    }
    return bazelOpts.workspaceName + '/' + result;
  };
```

which through `ngHost.amdModuleName` ends up in the rules_typescript tsc-wrapped code here:

```
  amdModuleName(sf: ts.SourceFile): string|undefined {
    if (!this.shouldNameModule(sf.fileName)) return undefined;
    // /build/work/bazel-out/local-fastbuild/bin/path/to/file.ts
    // -> path/to/file
    let fileName = this.rootDirsRelative(sf.fileName).replace(TS_EXT, '');

    let workspace = this.bazelOpts.workspaceName;

    // Workaround https://github.com/bazelbuild/bazel/issues/1262
    //
    // When the file comes from an external bazel repository,
    // and TypeScript resolves runfiles symlinks, then the path will look like
    // output_base/execroot/local_repo/external/another_repo/foo/bar
    // We want to name such a module "another_repo/foo/bar" just as it would be
    // named by code in that repository.
    // As a workaround, check for the /external/ path segment, and fix up the
    // workspace name to be the name of the external repository.
    if (fileName.startsWith('external/')) {
      const parts = fileName.split('/');
      workspace = parts[1];
      fileName = parts.slice(2).join('/');
    }

    if (this.bazelOpts.moduleName) {
      const relativeFileName = path.posix.relative(this.bazelOpts.package, fileName);
      if (!relativeFileName.startsWith('..')) {
        if (this.bazelOpts.moduleRoot &&
            this.bazelOpts.moduleRoot.replace(TS_EXT, '') ===
                relativeFileName) {
          return this.bazelOpts.moduleName;
        }
        // Support the common case of commonjs convention that index is the
        // default module in a directory.
        // This makes our module naming scheme more conventional and lets users
        // refer to modules with the natural name they're used to.
        if (relativeFileName === 'index') {
          return this.bazelOpts.moduleName;
        }
        return path.posix.join(this.bazelOpts.moduleName, relativeFileName);
      }
    }

    // path/to/file ->
    // myWorkspace/path/to/file
    return path.posix.join(workspace, fileName);
  }
```

which only knows how to convert a filename from the current compilation unit (`@angular/common/http`) to its module name. The best that this function can do with the filename `execroot/bazel_integration_test/bazel-out/darwin-fastbuild/bin/external/angular/packages/common/src/dom_tokens.ngfactory.d.ts` is return the module name `angular/packages/common/src/dom_tokens.ngfactory` which is incorrect.

This PR appends the logic of amdModuleName function to be able to handle converting filenames to module names for other modules besides the current compilation unit.

From the analysis above, the rational for why this is needed is because of the way imports for .ngfactory files are generated with full file paths that need to be converted to their module names for other modules besides the current one being compiled.

PR #226, opened 2018-06-28
Amd workaround in external ts_library rules should only affect TS <= 2.8

evmar:
b


evmar:
I solemnly swear that the extra commits are from Googlers.

gregmagolan:
For reference, the workaround is no longer needed as of tsc 2.9 as https://github.com/Microsoft/TypeScript/issues/18648 has been resolved for that version

alexeagle:
LGTM

Since rules_typescript uses Piper as source-of-truth, you should commit the CL first, with a `Closes #xx` tag in the CL description, then when you sync it will close this PR

PR #234, opened 2018-07-23
nodejs_binaries tsc and tsc_wrapped_bin should not depend on @build_bazel_rules_typescript_tsc_wrapped_deps//:node_modules

gregmagolan:
These binaries should depend on the default user's `@//:node_modules` version of typescript


gregmagolan:
The downside is that the user’s node_modules will need protobufjs 5.0.0 as well as the appropriate tsutils to match their tsc version (since these would have come from tsc_wrapped_deps previously).

If rules_typescript is pulled in via npm in a user's package.json then these deps will be pulled in automatically. Some more work to node_modules resolution for nodejs_binary is needed to ensure that nested deps are resolved properly for this to work in the case when a user's package.json has conflicting versions of these deps.

PR #237, opened 2018-07-27
Updated to rules_nodejs 0.11.2

gregmagolan:


internal/e2e/default_tsconfig_test.js
```diff
@@ -24,20 +24,35 @@ const tmpdir = fs.mkdtempSync(path.join(os.tmpdir(), 'wksp'));
 const WORKSPACE_BOILERPLATE = `
 http_archive(
     name = "build_bazel_rules_nodejs",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.10.0.zip"],
-    strip_prefix = "rules_nodejs-0.10.0",
+    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.11.2.zip"],
```
alexeagle: what do you think about this test? I was really down on nested WORKSPACEs but you seem to be doing okay with them in rules_nodejs (outside context of this PR)

gregmagolan: the nested WORKSPACE seem necessary for some integration tests. testing with preserve_symlinks on and off requires them in rules_nodejs.

just looked at that test.. it looks like you missed building workspace `a` and verifying that the build failed.

PR #239, opened 2018-07-27
fix: amd module names inconsistent in windows

DevVersion:
* Currently when building a `ts_library` within Windows, the AMD module name includes backslash delimiters. For consistency, the amd module names should always use posix paths.

cc. @alexeagle

### DIFF
--- a/internal/tsc_wrapped/compiler_host.ts
+++ b/internal/tsc_wrapped/compiler_host.ts
@@ -7,6 +7,7 @@ import {FileLoader} from './file_cache';
 import * as perfTrace from './perf_trace';
 import {BazelOptions} from './tsconfig';
 import {DEBUG, debug} from './worker';
+import {convertToPosixPath} from './posix-paths';
 
 export type ModuleResolver =
     (moduleName: string, containingFile: string,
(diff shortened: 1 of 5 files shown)


internal/tsc_wrapped/compiler_host.ts
```diff
@@ -205,8 +206,9 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
     for (const root of this.options.rootDirs) {
       if (fileName.startsWith(root)) {
         // rootDirs are sorted longest-first, so short-circuit the iteration
-        // see tsconfig.ts.
-        return path.posix.relative(root, fileName);
+        // see tsconfig.ts. Since the fileName can already include multiple path segments with
+        // backslash delimiters, we need to manually convert the path to posix format.
```
alexeagle: hmm, I'm surprised that fileName here could have backslashes, typescript should have canonicalized them already.
would you mind checking where that is coming from? it seems like it should have been converted somewhere else, this might be masking that bug

Also maybe it's time to have subtypes of `string` like `posixAbsolutePath` so the type checker could catch these, paths are such a mess :(

internal/tsc_wrapped/posix-paths.ts
```diff
@@ -0,0 +1,21 @@
+import {resolve} from 'path';
+
+/**
+ * Converts a path which could potentially include backslash delimiters into a consistent posix
+ * path that uses forward slashes.
+ * 
+ * Note that path.posix.resolve('a\\b\\c') still returns a non-posix path with backslashes.
+ */
+export function convertToPosixPath(value: string): string {
```
alexeagle: could you just leave the code in `tsconfig.ts`?

internal/tsc_wrapped/compiler_host.ts
DevVersion: Actually, it's my fault. I still ran an older version of the rules locally that didn't include https://github.com/bazelbuild/rules_typescript/commit/c52f257b6e33adc004ebdf0d573d3ec4e0c75c13. 

Commit https://github.com/bazelbuild/rules_typescript/commit/c52f257b6e33adc004ebdf0d573d3ec4e0c75c13 fixes the issue and there is no need to keep this PR open anymore. Thanks for the help.

DevVersion:
Actually I need to excuse.

PR #240, opened 2018-07-31
Generate BUILD.bazel files before running tests

fweikert:
This should generate all the necessary files by running the "//:gazelle" target before testing anything.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -2,6 +2,7 @@
 platforms:
   ubuntu1404:
     run_targets:
+    - "//:gazelle"
     - "@nodejs//:yarn"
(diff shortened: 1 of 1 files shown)


fweikert:
Is it expected that Gazelle does not work on Windows?

alexeagle:
Yes, see the comment right next to your edit about how rules_go is broken - https://github.com/bazelbuild/rules_typescript/pull/240/files#diff-b87fc4746c05e7b451e025d6d2914276R40

I think we'll just need to exclude all the go packages from the windows CI right now.
/cc @jayconrod

alexeagle:
was merged in 312bfeb8546997d67034e93aaedbc0906e13fefc

PR #241, opened 2018-07-31
Check rules_nodejs version, add check_rules_typescript_version

gregmagolan:


version.sh
```diff
@@ -0,0 +1,4 @@
+PACKAGE_VERSION=$(cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[\",]//g' | tr -d '[[:space:]]')
+sed 's/VERSION \= \"[0-9\.]*\"/VERSION = \"'${PACKAGE_VERSION}'\"/g' package.bzl > package.bzl.new
```
alexeagle: use `sed -i` to do the replacement inline so you don't need tmp file?

version.sh
```diff
@@ -0,0 +1,4 @@
+PACKAGE_VERSION=$(cat package.json | grep version | head -1 | awk -F: '{ print $2 }' | sed 's/[\",]//g' | tr -d '[[:space:]]')
```
alexeagle: what about writing `version_fix.js` and then you can just `require('package.json').version` ?

package.bzl
```diff
@@ -0,0 +1,45 @@
...
+"""Package file which defines build_bazel_rules_typescript version in skylark
+
+check_rules_typescript_version can be used in downstream WORKSPACES to check
+against a minimum dependent build_bazel_rules_typescript version.
+"""
+
+load("@build_bazel_rules_nodejs//internal/common:check_version.bzl", "check_version")
+
+# This version needs to be synced with the version in package.json
+# TODO(gmagolan): remove this duplication in the future for the release process
```
alexeagle: don't need the TODO anymore, right?

version.sh
gregmagolan: thought about that and even tried it but it looks like `-i` is a standard sed option between distros and it varies in how its used... safer to just do the `mv`

version.sh
gregmagolan: yeah.. that would work.. but then execute `sed` from within node or use node file i/o to update `package.bzl`?

package.bzl
gregmagolan: 👍

PR #244, opened 2018-08-02
Allow users to set the googmodule bazelOpt in their tsconfig

### LINKED ISSUE #230: Add support for generating goog modules
Currently the [`tsc_wrapped/compiler_host.ts`](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/compiler_host.ts) reads the [`googmodule` property](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/compiler_host.ts#L113) from the tsconfig, but it isn't ever set in [tsconfig.bzl](https://github.com/bazelbuild/rules_typescript/blob/master/internal/common/tsconfig.bzl#L122).

In my testing, setting it to `true` causes the [`ngc_wrapped/index.ts`](https://github.com/angular/bazel-builds/blob/master/src/ngc-wrapped/index.ts) compiler to output goog modules, because it uses the `tsc_wrapped/compiler_host.ts` and [emits with tsickle](https://github.com/angular/bazel-builds/blob/master/src/ngc-wrapped/index.ts#L260), but I haven't been able to get [`tsc_wrapped.ts`](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/tsc_wrapped.ts) to emit goog modules directly because it appears to use [some other emit](https://github.com/bazelbuild/rules_typescript/blob/master/internal/tsc_wrapped/tsc_wrapped.ts#L145).

alexeagle:
Fixes #230

### DIFF
--- a/internal/tsc_wrapped/tsc_wrapped.ts
+++ b/internal/tsc_wrapped/tsc_wrapped.ts
@@ -1,4 +1,5 @@
 import * as path from 'path';
+import * as tsickle from 'tsickle';
 import * as ts from 'typescript';
 
 import {PLUGIN as tsetsePlugin} from '../tsetse/runner';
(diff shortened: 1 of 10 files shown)


JaredNeil:
This should probably also set `bazelOpts.tsickle` to `true`, or also let the `tsickle` option to be set from the `tsconfig.json`.

internal/tsc_wrapped/tsc_wrapped.ts
```diff
@@ -139,15 +140,32 @@ function runOneBuild(
     return false;
   }
 
+  const emitter = bazelOpts.googmodule || bazelOpts.tsickle ? 'tsickle' : 'ts';
```
mprobst: wouldn't this be easier as a plain boolean `emitWithTsickle`?

internal/tsc_wrapped/tsc_wrapped.ts
```diff
@@ -139,15 +140,32 @@ function runOneBuild(
...
+  const emitResults: ts.EmitResult[] = [];
```
mprobst: maybe make this a `ts.EmitResult[]|tsickle.EmitResult[]`?

alexeagle: Doesn't help much, still need to cast it below to narrow out the ts.EmitResult which is missing the `modulesManifest`

PR #246, opened 2018-08-03
Create .bazelrc

cvcal:
See bazelbuild/bazel#5756


alexeagle:
This repo has source-of-truth in Piper, may as well just make the change as a google3 CL.

PR #249, opened 2018-08-08
Fix issue with module_root when building external repo

alexeagle:
Closes #242


pierreis:
Do we have any update on merging this?

alexeagle:
Hilarious, I just ran across this bug and debugged it again, finally discovered that this isn't merged :(

PR #250, opened 2018-08-09
Add CI config to test on RBE.

xingao267:


xingao267:
It needs bzip2 in the container. We'll hold on a little on this PR until there is an appropriate container.

PR #253, opened 2018-08-13
Allow users to disable auto quoting in tsickle

JaredNeil:
tsickle's auto-quoting feature sometimes causes unexpected behavior (see https://github.com/angular/tsickle/issues/683). tsickle accepts a flag to disable this feature, and this PR exposes that flag to the user.

### DIFF
--- a/examples/googmodule/a.ts
+++ b/examples/googmodule/a.ts
@@ -1,1 +1,6 @@
 export const a: number = 1;
+interface Quoted {
+  [k: string]: number;
+}
+let quoted: Quoted = {};
+quoted.hello = a;
--- a/examples/googmodule/googmodule_output_test.js
+++ b/examples/googmodule/googmodule_output_test.js
@@ -3,18 +3,18 @@ const fs = require('fs');
 describe('googmodule', () => {
   let output;
   beforeAll(() => {
-    output = require.resolve(
-        'build_bazel_rules_typescript/examples/googmodule/a.js');
+    output = fs.readFileSync(require.resolve(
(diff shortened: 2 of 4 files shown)


alexeagle:
Sorry, lost track of this - please ping me if you need a change merged, my github notifications are out of control.
We don't know of any user who wants tsickle auto-quoting under Bazel. I think I'd prefer to keep the configuration space small and just turn it off for everyone, what do you think?

JaredNeil:
@alexeagle, that sounds good to me. I've updated the PR to just always disable auto quoting.

alexeagle:
I pushed another commit to make the change only affect Bazel users (not changing the default for Blaze) and added a test

JaredNeil:
:+1: 
LGTM

evmar:
I think the proper place for this default is in tsickle, not here.

JaredNeil:
@evmar, there is a default in tsickle. It's just the opposite of the default we want here.

evmar:
Right, I'm saying if it makes sense to change the default, we should just change the default.

alexeagle:
Discussed with Evan, we think `disableAutoQuoting` got it backwards, it should be opt-in in tsickle.

After https://github.com/angular/tsickle/pull/895 is merged and released, you can update your tsickle version and no change to rules_typescript is required.

alexeagle:
new `--enableAutoQuoting` behavior shipped in tsickle 0.33.0
https://unpkg.com/tsickle@0.33.0/src/main.js
so this is now obsolete

PR #256, opened 2018-08-14
Move dependencies to package.bzl

alexeagle:
Simplifies user install instructions and puts us in control of versions of transitive deps

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -13,21 +13,15 @@
 # limitations under the License.
 
 workspace(name = "build_bazel_rules_typescript")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-http_archive(
-    name = "build_bazel_rules_nodejs",
-    urls = ["https://github.com/bazelbuild/rules_nodejs/archive/0.11.3.zip"],
-    strip_prefix = "rules_nodejs-0.11.3",
-    sha256 = "e8842fa5f5e38f2c826167ff94323d4b5aabd13217cee867d971d6f860cfd730"
+load(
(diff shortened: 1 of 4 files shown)


WORKSPACE
```diff
@@ -13,22 +13,22 @@
 # limitations under the License.
 
 workspace(name = "build_bazel_rules_typescript")
-load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
 
-http_archive(
+# TEMPORARY: grab rules_nodejs from Alex fork
```
filipesilva: Needs to be changed after https://github.com/bazelbuild/rules_nodejs/pull/279 is merged

alexeagle: done

defs.bzl
```diff
@@ -24,7 +24,8 @@ load("//internal/karma:ts_web_test.bzl",
      _ts_web_test = "ts_web_test_macro",
      _ts_web_test_suite = "ts_web_test_suite")
 load("//internal/protobufjs:ts_proto_library.bzl", _ts_proto_library = "ts_proto_library")
-load("//:package.bzl", _check_rules_typescript_version = "check_rules_typescript_version")
+load("//:package.bzl", "VERSION")
```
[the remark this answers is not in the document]
alexeagle: yeah, the difference is that rules_nodejs doesn't need an external repo to load the check function, so it can be together. in rules_typescript we can't have a load statement in the package.bzl file since it's responsible for fetching the rules_nodejs repo.

We could change rules_nodejs to have the check function outside the package.bzl.

PR #259, opened 2018-08-20
Override resolveTypeReferenceDirectives in compiler_host.ts to fix secondary type reference directives resolve under Bazel

gregmagolan:


internal/tsc_wrapped/compiler_host.ts
```diff
@@ -342,6 +343,64 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
...
+        }
+      });
+
+      // secondary search
+      if (!resolved) {
+        resolved = this.resolveTypingFromDirectory(path.posix.join(this.bazelOpts.nodeModulesPrefix, name), false);
+      }
+
+      if (!resolved) {
+        console.error(`Failed to resolve type reference directive '${name}'`);
```
gregmagolan: Could throw here... but typescript will fail silently if a reference type directive is not found so question is if we want to keep compiling or give up with an error?

alexeagle: I think we want to have behavior match TypeScript as much as possible.

internal/e2e/reference_types_directive/reference_types.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+// for this compilation includes "types": ["jasmine"].
+// This file will fail to compile if that mechanism of including types
+// is broken
+
+describe('reference types directive resolution', () => {
+  it('should resolve zone.js types from node_modules/zone.js/dist/zone.js.d.ts', () => {
+    // The type of Zone should resolve or this will fail to compile
+    // but since zone is not part of type test, typeof Zone should be
+    // equal to 'undefined'
+    expect(typeof Zone).toEqual('undefined');
```
alexeagle: it's pretty confusing - it looks like we're asserting that zone is broken, despite your comment.
Can we have any positive test first? Even just
let zone: Zone;
or something that would fail if the reference weren't there?

gregmagolan: good call. `let zone: Zone;` is less confusing

PR #260, opened 2018-08-22
Remove dependency on tsickle

alexeagle:
There is still a compile-time dependency, so to avoid users having to install tsickle,
this depends on shipping a JS distro of @bazel/typescript on npm


alexeagle:
I rewrote the commit to catch exceptions from `require('tsickle')` to tell the user what to do, PTAL

PR #261, opened 2018-08-27
WIP: feat(cypress_test): introduce rule

Toxicable:
Prototype of rule for running intergration tests with Cypress.

Right now it's only for `cypress run` (CI version) since there's another issue (see below) with `cypress open` (dev version)
Also it only works with `bazel run` since with bazel test due to the issues with bazel wanting to write files (see below)

These two issues are blocking the use of `bazel test`
https://github.com/cypress-io/cypress/issues/1281

This one is blocking `cypress open`
https://github.com/cypress-io/cypress/issues/1925

### DIFF
--- a/defs.bzl
+++ b/defs.bzl
@@ -24,6 +24,7 @@ load("//internal/karma:ts_web_test.bzl",
      _ts_web_test = "ts_web_test_macro",
      _ts_web_test_suite = "ts_web_test_suite")
 load("//internal/protobufjs:ts_proto_library.bzl", _ts_proto_library = "ts_proto_library")
+load("//internal/cypress_test:cypress_test.bzl", _cypress_test = "cypress_test")
 
 ts_setup_workspace = _ts_setup_workspace
 ts_library = _ts_library
@@ -35,3 +36,4 @@ ts_web_test_suite = _ts_web_test_suite
 ts_proto_library = _ts_proto_library
 # DO NOT ADD MORE rules here unless they appear in the generated docsite.
 # Run yarn skydoc to re-generate the docsite.
+cypress_test = _cypress_test
--- a/examples/integration/BUILD.bazel
+++ b/examples/integration/BUILD.bazel
@@ -1,0 +1,12 @@
+load("//:defs.bzl", "ts_library", "cypress_test")
+
+ts_library(
+    name = "test_lib",
+    srcs = [":index.spec.ts", "another.spec.ts"],
+)
+
+cypress_test(
+    name = "test",
+    deps = [":test_lib"],
+    srcs = []
+)
--- a/examples/integration/BUILD.bazel
+++ b/examples/integration/BUILD.bazel
@@ -2,7 +2,7 @@ load("//:defs.bzl", "ts_library", "cypress_test")
 
 ts_library(
     name = "test_lib",
-    srcs = [":index.spec.ts", "another.spec.ts"],
+    srcs = [":index.spec.ts"],
 )
 
 cypress_test(
--- a/examples/integration/another.spec.ts
+++ b/examples/integration/another.spec.ts
@@ -1,5 +1,0 @@
-describe('My First Test', function() {
-  it('Does not do much!', function() {
-    (expect(true) as any).to.equal(true)
-  })
-})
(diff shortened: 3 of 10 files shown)


Globegitter:
@Toxicable this seems quite exciting also for us, is there a reason this is in rules_typescript rather than rules_nodejs? As in this could also just be used with normal js, right?

Toxicable:
@Globegitter I wanted to ensure this worked with TS, but you're right it should be in rules_nodejs.
Since we're blocked by Cypress on a few issues for now I just wanted to show it somewhat working.

Toxicable:
So I tried doing a bit more work on this but ran into a few road blocks.

I don't know how to write a file (cypress.json) as part of of the test setup, right now you can see that im using `fs.writeFile` which will throw if used with `bazel test`.
I was able to add the file as part of the manifest
```
  all_data += [
    Label("//tools/cypress-rule:cypress_runner.js"),
    Label("//tools/cypress-rule:cypress.json"),
  ]
```
But the path that bazel provides me with is back to the source file, but I need the cypress.json to be copied into the directory with the test files aka: Execute root, I don't know how to get this file there.
Some advise on how to get around this would be appciated.

To work around the Cypress FS issues I was planning on using a fork in the meanwhile

Toxicable:
@Globegitter Thanks for the resources!
So looks like `TEST_UNDECLARED_OUTPUTS_DIR` is a directory under the exe root unfortunately, meaning I cant put the cypress.json there either.

I think a custom rule is going to have to be the way to proceed here

Globegitter:
@Toxicable I see you are making some progress with the native rule. Is it working now? OR at least looking more promising?

Toxicable:
@Globegitter I made some progress; this is my latest: https://github.com/bazelbuild/rules_typescript/blob/e664c44ae372e0efd8b758b930364ff318081f49/internal/cypress_test/cypress_test.bzl
But im hitting two issues at the moment
- The `cypress_bin` runfiles are not being added to the runfiles for my rule, therefore the start script for `cypress_bin` are there but `node` isn't
- only the `.d.ts` files are there, not the `.js` (edited)
Unfortunately I cant use fine grained deps to do this just yet since our repo isn't upgraded due to some issues with debugging nodejs code, however it should still be doable without it.

Globegitter:
@Toxicable yeah runfiles can sometimes be a bit tricky, I commented on an issue related to that once (can't find it anymore) but the api is a bit confusing and redundant. I will try and make some time to test this out myself in the coming days. Have you tried adding it to `transitive_files` maybe?

Toxicable:
@Globegitter No luck on that one just yet.
I found that if I used ctx.actions.run instead of DefaultInfo I can execute the cypress binary without issues, but I don't see how to make that work with `bazel test` properly since I have no outputs, I havn't found any examples doing this either.
It would appear the only way to do tests is with a script + DefaultInfo

For the first issue:
I have both these set: 
```
          collect_data = True,
          collect_default = True,
```
which do: https://docs.bazel.build/versions/master/skylark/lib/ctx.html#runfiles
>Whether to collect the default runfiles from the dependencies in srcs, data and deps attributes.

But the cypress executable is not in srcs, data or deps, nor do I see how I'm meant to add cypress to those attriibutes or how it's done in the karma rule im looking at.

For the second:
I think this is whats meant to collect up the .js files
```
  files = depset(ctx.files.srcs)
  for d in ctx.attr.deps:
    if hasattr(d, "node_sources"):
      files += d.node_sources
    elif hasattr(d, "files"):
      files += d.files
```
which I can see when iterating it
`d` in my case is a `<target //tools/cypress-rule/test:test_lib, keys:[OutputGroupInfo]>` 
that only has `d.files` which is: `depset([<generated file tools/cypress-rule/test/index.spec.d.ts>])`
and that index.spec.d.ts is in the right place, but I don't know where it's js files are at

I can see here: https://github.com/bazelbuild/rules_typescript/blob/master/internal/karma/ts_web_test.bzl#L82
That this `expand_path_into_runfiles` function is used which may resolve the .d.ts into .js (?), but `user_entries` is never added to the runfiles so I don't think it grabs those files.

alexeagle:
@Toxicable I'm going to close this PR now that rules_nodejs is a monorepo with packages/ subdirs. Want to introduce the cypress subdirectory there?

Toxicable:
I've opened a ticket in the wrules_nodejs repo: https://github.com/bazelbuild/rules_nodejs/issues/607
To track the issues with implementing this rule

PR #264, opened 2018-09-06
Tell users when their version override works

alexeagle:


kyliau:
@alexeagle ok to close this? Doesn't look like a proper change.

PR #266, opened 2018-09-10
Allow setup files in ts_web_test_suite

filipesilva:


filipesilva:
Well that answers what I was asking @gregmagolan about how to avoid circular deps with angular 👍

internal/karma/karma.conf.js
```diff
@@ -110,6 +110,29 @@ resolveRunfiles(manifest, [
...
+    const BrowserDynamicTestingModule = 
+      require('@angular/platform-browser-dynamic/testing').BrowserDynamicTestingModule;
+    const platformBrowserDynamicTesting = 
+      require('@angular/platform-browser-dynamic/testing').platformBrowserDynamicTesting;
+    
+    getTestBed().initTestEnvironment(
+      BrowserDynamicTestingModule,
+      platformBrowserDynamicTesting()
+    );
+  } catch (e) { }
```
Toxicable: even though this isn't meant to go in this repo a message here would be great such as:
`Unable to automatically initalize TestBed.
  You must manually initalize TestBed if you intend to use it in your tests.
  Error:
  ${stack tace / message}
`

filipesilva: Yeah we should probably put that in the example repo.

internal/karma/karma.conf.js
```diff
@@ -114,14 +115,15 @@ var requireConfigContent = `
 // A simplified version of Karma's requirejs.config.tpl.js for use with Karma under Bazel.
 // This does an explicit \`require\` on each test script in the files, otherwise nothing will be loaded.
 (function(){
+  var setupFile = [TMPL_setup_file].map(function(file) { return file.replace(/\\.js$/, ''); });
```
gregmagolan: setupFiles seems like

internal/karma/ts_web_test.bzl
```diff
@@ -167,6 +173,10 @@ ts_web_test = rule(
             allow_files = True,
             aspects = [sources_aspect],
         ),
+        "setup_files": attr.label_list(
```
gregmagolan: it feels strange to accept .js files here. it makes it seem like these should user .js files that should be added to the runfiles instead of just filenames that should match the output of a ts_library rule.

gregmagolan: bootstrap accepts .js files but also adds them to runfiles like so `karma_runfiles += ctx.files.bootstrap` which is what you'd expect when you accept a filename

filipesilva: It's definitely an array of entry points, yeah. 

I'm not sure about the label approach though. A label can have many files in it. How do we know which one is the entrypoint then?

alexeagle: we error at runtime if you provide a label that is more than one file.
I think this is idiomatic in bazel (currently) - discussed a design with the Bazel TL that does that.

internal/karma/ts_web_test.bzl
```diff
@@ -78,6 +78,15 @@ def _ts_web_test_impl(ctx):
         for f in ctx.files.static_files
     ]
 
+    setup_files = []
+    for d in ctx.attr.setup_deps:
+        if not hasattr(d, "typescript"):
+            fail("labels in setup_deps must be created by ts_library")            
+        es5_sources = d.typescript.es5_sources.to_list()
+        if len(es5_sources) != 1:
+            fail("labels in setup_deps must contain exactly one file")
```
gregmagolan: it may be better to allow multiple files per label as not all setup steps have to be ordered so it would be less overhead for a user with multiple setup steps to be able to use a single ts_library with a glob

internal/karma/ts_web_test.bzl
```diff
@@ -78,6 +78,15 @@ def _ts_web_test_impl(ctx):
         for f in ctx.files.static_files
     ]
 
+    setup_files = []
+    for d in ctx.attr.setup_deps:
+        if not hasattr(d, "typescript"):
```
gregmagolan: You're checking if the setup_dep has a typescript provider so that you can get at the es5_sources? Seems like an overly strict requirement since it prevents a user from adding a .js file to setup_files. Maybe this should follow the same pattern as user_entries:
```
setup_deps = depset([])
for d in ctx.attr.setup_deps:
    if hasattr(d, "node_sources"):
        setup_deps += d.node_sources
    elif hasattr(d, "files"):
        setup_deps += d.files
setup_files = [
    expand_path_into_runfiles(ctx, f.short_path)
    for f in setup_deps
]
```
The above doesn't check for the single file requirement but that could be done as well in the loop if it is desirable

internal/karma/ts_web_test.bzl
```diff
@@ -31,7 +31,7 @@ def _ts_web_test_impl(ctx):
     )
 
     files = depset(ctx.files.srcs)
-    for d in ctx.attr.deps:
+    for d in ctx.attr.deps + ctx.attr.setup_deps:
```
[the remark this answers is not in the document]
filipesilva: This is doing a bit more though, it's adding the node_sources as well which will be necessary for the setup files to run.

All the ts_library files in `TMPL_user_files` are being added as requirejs modules, so they have to be there before being required proper.

The testbed example should be something like this;
```
import 'zone.js/dist/zone-testing';
import { getTestBed } from '@angular/core/testing';
import {
  BrowserDynamicTestingModule,
  platformBrowserDynamicTesting
} from '@angular/platform-browser-dynamic/testing';

getTestBed().initTestEnvironment(
  BrowserDynamicTestingModule,
  platformBrowserDynamicTesting()
);
```

Even though this is a single setup file, it depends on three libraries. We only want this file to be required as an entry point, but all the libraries must be available to be required.

internal/karma/ts_web_test.bzl
filipesilva: I think it makes sense to also allow js_library files, but I don't think it will work with generic JS files since trying to `require()` them will fail.

Is it more idiomatic in bazel to guard against unexpected inputs, or to be permissive?

internal/karma/ts_web_test.bzl
filipesilva: I'm fine either way, it just depends on what's the most idiomatic approach in Bazel. 

In the more permissive scenario (any number of files per label) it's up to the user to not shoot themselves in the foot since they are all effectively entry points and the order of execution is undefined.

I'm not sure there are a lot of cases beyond super trivial ones where the order doesn't matter though. I guess we can chalk it up to user fault if they have a bunch of modules that do stuff at the toplevel.

gregmagolan: Good question about what is more idiomatic. I'm in favor of allowing multiple inputs and documenting that the order of execution of the files within each label is not deterministic. @alexeagle What do you think?

alexeagle: Looking at this with @kyliau 

Maybe it would be better to use the same pattern from `nodejs_binary`.

1. list these extra deps as normal in the `deps`. they won't be executed since they don't match the filename filter for specs
2. the setup argument would be a list of strings, each of which is to be `require`d


"bootstrap | List of strings; Optional; Default is []JavaScript modules to be loaded before the entry point. For example, Angular uses this to patch the Jasmine async primitives for zone.js before the first describe."

alexeagle:
@kyliau and I spent a bunch more time today digging into this.
1) in google3 we solve for this use case with a rule `ts_development_sources` whose job is to extract the sources from upstream `ts_library` rules. `ts_development_sources` takes an attribute `runtime_deps` which is a list of labels whose files are loaded first, before the `srcs` and `deps`. so we'd like to rename `setup_deps` to `runtime_deps` to make this match
2) we think it's fine to load all the files produced by the target whose label is provided. sure, it's up to the user to list those in the appropriate order, if there is an order dependency between them
3) there's a file here `manifest.ts` under `tsc_wrapped` which actually produces the import-topologically-sorted manifest of JS files, which we use in google3 to get the right load order. Oops, that's not even used under Bazel yet

If you could rebase and make the changes for 1 and 2, I'll do some cleanup of docs tomorrow morning first thing and get this in.

filipesilva:
@alexeagle @kyliau made the requested changed, PTAL.

PR #269, opened 2018-09-14
Expand Windows test coverage

filipesilva:
Fix https://github.com/bazelbuild/rules_typescript/issues/270
Fix https://github.com/bazelbuild/rules_typescript/issues/272

### DIFF
--- a/internal/karma/karma.conf.js
+++ b/internal/karma/karma.conf.js
@@ -43,13 +43,13 @@ if (process.env['WEB_TEST_METADATA']) {
     if (webTestNamedFiles['CHROMIUM']) {
       // When karma is configured to use Chrome it will look for a CHROME_BIN
       // environment variable.
-      process.env.CHROME_BIN = path.join('external', webTestNamedFiles['CHROMIUM']);
+      process.env.CHROME_BIN = require.resolve(webTestNamedFiles['CHROMIUM']);
       browsers.push(process.env['DISPLAY'] ? 'Chrome': 'ChromeHeadless');
     }
     if (webTestNamedFiles['FIREFOX']) {
       // When karma is configured to use Firefox it will look for a FIREFOX_BIN
       // environment variable.
-      process.env.FIREFOX_BIN = path.join('external', webTestNamedFiles['FIREFOX']);
+      process.env.FIREFOX_BIN = require.resolve(webTestNamedFiles['FIREFOX']);
       browsers.push(process.env['DISPLAY'] ? 'Firefox': 'FirefoxHeadless');
     }
   } else {
@@ -64,6 +64,7 @@ if (!browsers.length) {
   browsers.push(process.env['DISPLAY'] ? 'Chrome': 'ChromeHeadless');
 }
 
+const proxies = {};
 const files = [
   TMPL_bootstrap_files
   TMPL_user_files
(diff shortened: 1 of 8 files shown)


package.bzl
```diff
@@ -47,9 +47,9 @@ def rules_typescript_dependencies():
     _maybe(
         http_archive,
         name = "io_bazel_rules_webtesting",
-        urls = ["https://github.com/bazelbuild/rules_webtesting/archive/89219129b13ec0f2cd6e65eff9113948e84e1856.zip"],
-        strip_prefix = "rules_webtesting-89219129b13ec0f2cd6e65eff9113948e84e1856",
-        sha256 = "71bcc2dc0d1d9c9e4f032dd2c0983fa73cb90adb48e86453f78071af8b2fc7ec",
+        urls = ["https://github.com/filipesilva/rules_webtesting/archive/d459ccfe31478bdddc71afe63dd9688399382ea0.zip"],
```
filipesilva: This PR is using my forked version of rules_webtesting to test the chrome fix. Needs to be changed back to the canonical repo once https://github.com/bazelbuild/rules_webtesting/pull/303 is merged.

internal/karma/karma.conf.js
```diff
@@ -4,6 +4,86 @@ const path = require('path');
...
+const WORKSPACE_NAME = 'TMPL_workspace_name';
+
+// On Windows, runfiles will not be in the runfiles folder but instead
+// there is a MANIFEST file which maps the runfiles for the test
+// to their location on disk. Bazel provides a RUNFILE_MANIFEST_FILE environment
+// variable which is set the location of the MANIFEST file during test execution.
+// If a MANIFEST file is found, we remap the test files to their
+// location on disk using the MANIFEST file.
+let manifest = null;
+if (process.env.RUNFILES_MANIFEST_ONLY === '1' &&
```
alexeagle: we should throw an error if RUNFILES_MANIFEST_ONLY=1 and the file doesn't exist ?

internal/karma/karma.conf.js
```diff
@@ -4,6 +4,86 @@ const path = require('path');
...
+
+function readDir(dir) {
+  return fs.statSync(dir).isDirectory() ?
+    Array.prototype.concat(...fs.readdirSync(dir).map(f => readDir(path.join(dir, f)))) :
+    dir.replace(/\\/g, '/');
+}
+
+function resolveManifestFile(runfilesManifest, res) {
+  const maybe = runfilesManifest[res];
+  if (maybe) {
```
alexeagle: can we use the runfiles JS library support rather than copy-paste code? @gregmagolan  can help

internal/ts_repositories.bzl
```diff
@@ -34,6 +34,12 @@ def ts_setup_workspace():
         importpath = "github.com/kylelemons/godebug",
     )
 
+    go_repository(
```
alexeagle: I wonder how this will work in google3... I'll probably have to do some stitching...

internal/karma/karma.conf.js
filipesilva: I didn't know we had one, and was going to suggest making it. Yeah I think we should, and we should use it throughout everywhere. It needs to be able to look up runfiles generally and not just node resolution.

alexeagle: actually it seems like we should land https://github.com/bazelbuild/rules_typescript/pull/278 first

PR #273, opened 2018-09-15
Update to rules_nodejs 0.13.2

gregmagolan:


gregmagolan:
Waiting on rules_nodejs 0.13.2 tag to be pushed

PR #276, opened 2018-09-18
Idiomatic install

gregmagolan:


BUILD.bazel
```diff
@@ -34,19 +36,26 @@ gazelle(
...
-    # Only include files needed for type-checking and runtime
-    srcs = glob([
-        "node_modules/**/*.js",
-        "node_modules/**/*.d.ts",
-        "node_modules/**/*.json",
-    ] + [
-        # Needed because http-server has a bin with no .js extension
-        "node_modules/http-server/**",
-    ]),
+    srcs = glob(
```
alexeagle: could we switch to bazel-managed deps now?

BUILD.bazel
```diff
@@ -66,3 +75,17 @@ js_library(
...
+    name = "npm_package",
+    srcs = [
+        "LICENSE",
+        "README.md",
+        "package.json",
+        "//tools:check_version.js",
+    ],
+    deps = [
+        "//internal:tsc_wrapped",
+        "//internal/karma:karma_concat_js",
```
alexeagle: karma should be in its own npm package.
see my changes in https://github.com/alexeagle/rules_typescript/tree/distribute_js

internal/build_defs.bzl
```diff
@@ -198,7 +198,8 @@ ts_library = rule(
             Sets a different TypeScript compiler binary to use for this library.
             For example, we use the vanilla TypeScript tsc.js for bootstrapping,
             and Angular compilations can replace this with `ngc`.""",
-            default = Label("//internal:tsc_wrapped_bin"),
+            default = Label("@npm//:@bazel/typescript/tsc_wrapped"),
```
alexeagle: probably needs some docs added in case the user didn't name their `yarn_install` rule `npm`

internal/defaults.bzl
```diff
@@ -0,0 +1,22 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"Defaults for rules_typescript repository not meant to be used downstream"
+
+load("@build_bazel_rules_typescript//:defs.bzl", _ts_library = "ts_library")
+
+DEFAULT_TS_LIBRARY_COMPILER = "@build_bazel_rules_typescript//internal:tsc_wrapped_bin"
```
alexeagle: comment explaining that we can't use the default because that would be a JS dependency on npm, and we are building the thing itself to publish it to npm

internal/e2e/reference_types_directive/BUILD.bazel
```diff
@@ -21,6 +21,11 @@ ts_library(
 
 jasmine_node_test(
     name = "test",
-    deps = [":test_lib"],
-    node_modules = "//:node_modules",
+    deps = [
+      "@npm//:jasmine",
+      "@npm//:which",
+      # @npm//:which this is needed as the 'exit' package doesn't
```
alexeagle: at some point this comment will have to change so that ts_auto_deps understands not to prune the unused dependency.

We had also once discussed being able to add these undeclared dependencies in the `yarn_install` rule so the `@npm//:exit` package definition would get `which` added. Otherwise you have to duplicate this comment in many tests?

tools/check_version.js
```diff
@@ -0,0 +1,46 @@
+const path = require('path');
+const fs = require('fs');
+var {spawnSync} = require('child_process');
+
```
alexeagle: needs file-level comment

tools/check_version.js
```diff
@@ -0,0 +1,46 @@
...
+  contents = fs.readFileSync(maybe, 'utf8');
+}
+
+// local yarn call in build_bazel_rules_typescript WORKSPACE
+maybe = path.resolve(process.cwd(), 'package.bzl');
+if (fs.existsSync(maybe)) {
+  contents = fs.readFileSync(maybe, 'utf8');
+}
+
+if (!contents) {
```
alexeagle: logic might be easier to read if we comment about which failure mode we are checking for at each step

internal/build_defs.bzl
```diff
@@ -212,10 +218,10 @@ ts_library = rule(
         ),
         "tsickle_typed": attr.bool(default = True),
         "internal_testing_type_check_dependencies": attr.bool(default = False, doc = "Testing only, whether to type check inputs that aren't srcs."),
-        # @// is special syntax for the "main" repository
-        # The default assumes the user specified a target "node_modules" in their
-        # root BUILD file.
-        "node_modules": attr.label(default = Label("@//:node_modules")),
+        "node_modules": attr.label(
+            doc = """TODO(gregmagolan): doc string""",
+            default = Label("@npm//:typescript__dts"),
```
gregmagolan: This seems like the best way to avoid user's having to add `@npm//:typescript` as a dep to all `ts_library` targets since the default libs from the typescript package are always (almost always?) needed by ts_library. If a user isn't using bazel managed deps and doesn't have an `@npm` workspace then he'll have to override both the `compiler` and the `node_modules` attributes of ts_library and ng_module.

Here's the change needed in rules_nodejs to support this: https://github.com/bazelbuild/rules_nodejs/pull/328. We include only the dts files because we don't want all of the typescript .ts, .js & .json files being included as action inputs.

gregmagolan: I wonder if it would also make sense to include `node_modules/@types/**/*.d.ts` in this default node_modules filegroup since then all types would be available as would be default in typescript outside of bazel?

examples/BUILD.bazel
```diff
@@ -34,10 +36,13 @@ ts_library(
 ts_library(
     name = "bar_ts_library",
     srcs = ["bar.ts"],
-    tsconfig = ":tsconfig.json",
+    tsconfig = ":tsconfig-e2e.json",
```
gregmagolan: this should probably use `@npm//:@types` when it is available to show how all types can be automatically discovered. bar.ts has
```
// Repro for #31, should automatically discover @types/node
import * as fs from 'fs';
```

examples/app/BUILD.bazel
```diff
@@ -38,4 +39,10 @@ ts_library(
     name = "e2e",
     testonly = 1,
     srcs = ["app_e2e_test.ts"],
+    tsconfig = "//examples:tsconfig-e2e.json",
```
gregmagolan: `tsconfig-e2e.json` should be use the tsconfig_rule and extend from `tsconfig.json`

examples/googmodule/tsconfig.json
```diff
@@ -1,4 +1,10 @@
 {
+    "compilerOptions": {
+        // Don't auto-discover @types since we are on typescript 2.7 in and
```
gregmagolan: need to test if this is actually the case with ts 2.8, 2.9 and 3.0

examples/app/BUILD.bazel
gregmagolan: should probably rename tsconfig-e2e.json to tsconfig-test.json

internal/BUILD.bazel
```diff
@@ -70,10 +73,14 @@ ts_library(
 nodejs_binary(
     name = "tsc_wrapped_bin",
     data = [
+        "@npm//:protobufjs",
+        "@npm//:source-map-support",
+        "@npm//:tsickle",
```
gregmagolan: tsickle is required. hmmm.

internal/build_defs.bzl
```diff
@@ -45,6 +47,13 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file, node_opts, description
         for f in ctx.files.node_modules
         if f.path.endswith(".js") or f.path.endswith(".ts") or f.path.endswith(".json")
     ]
+    for d in ctx.attr.deps:
```
gregmagolan: need a comment here

internal/common/compilation.bzl
```diff
@@ -64,7 +66,7 @@ def assert_js_or_typescript_deps(ctx, deps = None):
     # Fallback to `ctx.attr.deps`.
     deps = deps or ctx.attr.deps
     for dep in deps:
-        if not hasattr(dep, "typescript") and not hasattr(dep, "js"):
+        if not hasattr(dep, "typescript") and not hasattr(dep, "js") and not NodeModuleInfo in dep:
```
gregmagolan: need a comment here and to update the fail msg

internal/e2e/default_tsconfig_test.js
```diff
@@ -60,12 +60,276 @@ local_repository(
     name = "build_bazel_rules_typescript",
     path = "${process.cwd()}",
 )
-load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories")
+load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "yarn_install")
 node_repositories(package_json=[])
```
gregmagolan: package_json defaults to [] so it doesn't need to explicitly set

internal/e2e/reference_types_directive/BUILD.bazel
gregmagolan: test if `@npm//:which` is still needed now that `node_modules` files are filtered out of `jasmine_node_test`

internal/karma/karma.conf.js
```diff
@@ -117,7 +117,7 @@ var requireConfigContent = `
   var allFiles = [TMPL_user_files];
   var allTestFiles = [];
   allFiles.forEach(function (file) {
-    if (/(spec|test)\\.js$/i.test(file) && !/\\/node_modules\\//.test(file)) {
+    if (/[^a-zA-Z0-9](spec|test)\\.js$/i.test(file) && !/\\/node_modules\\//.test(file)) {
```
gregmagolan: add word separator to filter (same as `jasmine_node_test`)

ts_auto_deps/main.go
```diff
@@ -14,6 +14,7 @@ var (
 		"(generates ts_config and ts_development_sources targets).")
 	recursive = flag.Bool("recursive", false, "recursively update all packages under the given root.")
 	files     = flag.Bool("files", false, "treats arguments as file names. Filters .ts files, then runs on their dirnames.")
+	internal  = flag.Bool("internal", false, "build files should be configured for tests in the rules_typescript repository.")
```
gregmagolan: changes to ts_auto_deps will be reverted

internal/e2e/ts_auto_deps/package.json
```diff
@@ -0,0 +1,6 @@
+{
+  "dependencies": {
+    "@bazel/typescript": "file:../build_bazel_rules_typescript/bazel-bin/internal/npm_package",
```
gregmagolan: Not sure if this will work on Windows. May have to disable tests that pull in the generated npm_package on Windows.

.circleci/bazel.rc
```diff
@@ -28,4 +28,4 @@ build --experimental_repository_cache=/home/circleci/bazel_repository_cache
 # Limit Bazel to consuming 2560K of RAM
 build --local_resources=2560,1.0,1.0
 # Also limit Bazel's own JVM heap to stay within our 4G container limit
-startup --host_jvm_args=-Xmx1g
+startup --host_jvm_args=-Xmx2g
```
alexeagle: maybe add a comment about how you measured the right value

BUILD.bazel
```diff
@@ -29,24 +29,33 @@ alias(
...
-    # Only include files needed for type-checking and runtime
-    srcs = glob([
-        "node_modules/**/*.js",
-        "node_modules/**/*.d.ts",
-        "node_modules/**/*.json",
-    ] + [
-        # Needed because http-server has a bin with no .js extension
-        "node_modules/http-server/**",
-    ]),
+    srcs = glob(
```
alexeagle: when will we switch to bazel-managed deps??

WORKSPACE
```diff
@@ -51,11 +51,27 @@ node_repositories(
   package_json = ["//:package.json"],
   preserve_symlinks = True)
 
+yarn_install(
+    name = "npm",
+    package_json = "//:package.json",
+    yarn_lock = "//:yarn.lock",
+    data = ["//tools:check_version.js"],
+    manual_build_file_contents = """
```
alexeagle: does this replace the generated BUILD file?

examples/BUILD.bazel
alexeagle: :+1:

internal/BUILD.bazel
alexeagle: maybe add a TODO, I'm not that worried about it

internal/BUILD.bazel
```diff
@@ -82,12 +89,47 @@ ts_library(
...
+)
+
+genrule(
+    name = "worker_protocol_copy",
+    srcs = ["//third_party/github.com/bazelbuild/bazel/src/main/protobuf:worker_protocol.proto"],
+    outs = ["worker_protocol.proto"],
+    cmd = "cp $< $@",
+)
+
+npm_package(
```
alexeagle: too bad this can't be `//internal/tsc_wrapped:npm_package` but I reached the same conclusion as you did

internal/karma/karma.conf.js
alexeagle: either way okay with me, this is close to landing

internal/karma/package.json
```diff
@@ -1,6 +1,14 @@
 {
+  "name": "@bazel/karma",
   "description": "runtime dependences for web test rules",
-  "devDependencies": {
+  "homepage": "https://github.com/bazelbuild/rules_typescript",
+  "license": "Apache-2.0",
+  "version": "0.0.0-PLACEHOLDER",
```
alexeagle: I think we need to add the stamping - can be separate PR (I could do it)

internal/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "name": "@bazel/typescript",
+  "description": "Custom TypeScript compiler for hosting on Bazel",
+  "homepage": "https://github.com/bazelbuild/rules_typescript",
+  "license": "Apache-2.0",
+  "version": "0.0.0-PLACEHOLDER",
+  "keywords": [
+      "typescript",
+      "bazel"
+  ],
+  "bin": {
```
alexeagle: maybe there's not much point advertising this as a bin
if you run it outside bazel it fails because of worker something
if you run it inside bazel you don't use this nodejs_binary

internal/tsc_wrapped/tsc_wrapped.ts
```diff
@@ -1,3 +1,4 @@
+#!/usr/bin/env node
```
alexeagle: if it's not a bin then we could drop this too?

BUILD.bazel
gregmagolan: done

tools/bazel_stamp_vars.sh
```diff
@@ -0,0 +1,2 @@
+#!/usr/bin/env bash
+echo BUILD_SCM_VERSION $(git describe --abbrev=7 --tags HEAD)
```
alexeagle: oh, you did it already, cool

internal/package.json
gregmagolan: agreed

internal/ts_repositories.bzl
```diff
@@ -43,24 +43,12 @@ def ts_setup_workspace():
     # 0.11.3: node module resolution fixes & check_rules_nodejs_version
     check_rules_nodejs_version("0.11.3")
 
-    yarn_install(
```
alexeagle: can't remove this yet because it's referenced in more places like https://github.com/ReactiveX/rxjs/blob/master/src/BUILD.bazel#L24

sorry that's my fault... :(

internal/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "@bazel/typescript",
+  "description": "Custom TypeScript compiler for hosting on Bazel",
+  "homepage": "https://github.com/bazelbuild/rules_typescript",
+  "license": "Apache-2.0",
+  "version": "0.0.0-PLACEHOLDER",
+  "keywords": [
```
alexeagle: needs a "typings" field so my downstream import of `@bazel/typescript` can work (no index.d.ts in the package)

tools/check_version.js
gregmagolan: done

tools/check_version.js
gregmagolan: cleaned this code up now that we have separate packages and check_version.js is not called in the main package.json anymore

internal/defaults.bzl
gregmagolan: done

internal/build_defs.bzl
gregmagolan: done

.circleci/bazel.rc
gregmagolan: done

internal/package.json
gregmagolan: done

internal/tsc_wrapped/tsc_wrapped.ts
gregmagolan: yup. done

BUILD.bazel
gregmagolan: done

examples/app/BUILD.bazel
gregmagolan: done

internal/BUILD.bazel
gregmagolan: kk

internal/common/compilation.bzl
gregmagolan: done

internal/package.json
gregmagolan: there is a .d.ts file for each compiled .ts file in the npm_package. do you mean to have a central index.d.ts then that would include all of them?

internal/build_defs.bzl
gregmagolan: An `@npm//:@types` scoped target is now generated that includes all of the types. I think its better to be explicit and not include these by default. User's that don't have `types = []` in their tsconfig can just depend on `@npm//:@types` to get all types into the compile.

examples/BUILD.bazel
gregmagolan: done

internal/build_defs.bzl
gregmagolan: done

internal/ts_repositories.bzl
gregmagolan: ok. its back in

examples/googmodule/tsconfig.json
gregmagolan: added tests for different typescript versions but still not sure under what conditions typescript will auto-discover types

internal/e2e/ts_auto_deps/package.json
gregmagolan: e2e tests not run on Windows

internal/package.json
gregmagolan: Added "typings": "./tsc_wrapped/index.d.ts" to @bazel/typescript package.json

internal/karma/karma.conf.js
```diff
@@ -39,19 +39,21 @@ if (process.env['WEB_TEST_METADATA']) {
...
-      process.env.CHROME_BIN = require.resolve(webTestNamedFiles['CHROMIUM']);
-      browsers.push(process.env['DISPLAY'] ? 'Chrome': 'ChromeHeadless');
-    }
-    if (webTestNamedFiles['FIREFOX']) {
-      // When karma is configured to use Firefox it will look for a FIREFOX_BIN
-      // environment variable.
-      process.env.FIREFOX_BIN = require.resolve(webTestNamedFiles['FIREFOX']);
-      browsers.push(process.env['DISPLAY'] ? 'Firefox': 'FirefoxHeadless');
-    }
+    webTestMetadata['webTestFiles'].forEach(webTestFiles => {
```
alexeagle: why did we look at only the first item before and now we look at all of them?
/cc @vikerman

gregmagolan: After updating to the latest rules_webtesting (Filipe’s Windows year coverage change), FIREFOX_BIN is no longer the first entry in the list.

Also, Firefox tests are broken as well after Filipe’s commit. Not sure how it passed CI in that PR.

gregmagolan: After updating to the latest rules_webtesting (Filipe’s Windows year coverage change), FIREFOX_BIN is no longer the first entry in the list.

Also, Firefox tests are broken as well after Filipe’s commit. Not sure how it passed CI in that PR.

PR #280, opened 2018-09-21
Re-enable gazelle on Windows

filipesilva:


filipesilva:
Blocked on https://github.com/bazelbuild/continuous-integration being updated to Bazel 0.17.1, see https://github.com/bazelbuild/bazel-gazelle/issues/321#issuecomment-423519381.

PR #282, opened 2018-09-24
Update skydoc repo to d9b3e3b, to fix skydoc #58

laszlocsomor:
As a consequence, Bazel on Windows can now build
//docs/...

See https://github.com/bazelbuild/skydoc/issues/58

### DIFF
--- a/package.bzl
+++ b/package.bzl
@@ -117,9 +117,9 @@ def rules_typescript_dev_dependencies():
 
     http_archive(
         name = "io_bazel_skydoc",
-        urls = ["https://github.com/bazelbuild/skydoc/archive/0ef7695c9d70084946a3e99b89ad5a99ede79580.zip"],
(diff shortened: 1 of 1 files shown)


laszlocsomor:
https://github.com/bazelbuild/skydoc/commit/d9b3e3b743f00bf00ba8d1306c56a477ec50fc4e is actually buggy, https://github.com/bazelbuild/skydoc/pull/109 has the fix. I'll update this PR once that one is merged.

laszlocsomor:
FYI @alexeagle

filipesilva:
I can confirm that locally I can build the `//docs/...` target, but we still will not be able to build it on CI because skydoc is not compatible with Python 3, which is the one used by https://github.com/bazelbuild/continuous-integration. 

There's some more context in https://github.com/bazelbuild/skydoc/issues/57#issuecomment-425110394.

laszlocsomor:
Thanks for the info!

PR #286, opened 2018-09-25
Examples workspace

filipesilva:


alexeagle:
We are going to re-do the TypeScript example in rules_nodejs now that we are moving user-facing parts of this repo there. I'm not sure if there's stuff in this PR that we should keep? I'll close it assuming it's just stale.

PR #289, opened 2018-09-27
Move io_bazel dep to devDeps

alexeagle:
I don't see why ts_auto_deps depends on io_bazel as is claimed in the comment


gregmagolan:
> I don't see why ts_auto_deps depends on io_bazel as is claimed in the comment

Hmmm. Not sure why it errored for me then if this PR is green. When I moved ts_auto_deps into a nested WORKSPACE I saw an error that suggested that but could have been mistaken.

gregmagolan:
Lint fix is this one: https://github.com/bazelbuild/rules_typescript/pull/288/files#diff-727f985f58b40703df0bf7e66d34b867R53

PR #291, opened 2018-09-28
Update to tsutils 2.27.2 in remaining locations

gregmagolan:


alexeagle:
Why wasn't anything red when I missed these? Doesn't the new tsetse require it? How did you discover it was broken?

gregmagolan:
> Why wasn't anything red when I missed these? Doesn't the new tsetse require it? How did you discover it was broken?

I'm not sure. It only showed up downstream in the angular repo so the e2e tests that are in this repo that pull in the built npm package must not exercise that code path

Issue #296, opened 2018-10-04
Karma firefox test failing with `Cannot find module 'Firefox.app/Contents/MacOS/firefox'`


gregmagolan:
After the last update of rules_webtesting in rules_typescript, `ts_web_test_suite` targets that test against the `@io_bazel_rules_webtesting//browsers:firefox-local` browser fail with `Cannot find module 'Firefox.app/Contents/MacOS/firefox'`.

This is true for OSX at least. I haven't checked if it fails on Linux or Windows.

//cc @filipesilva @alexeagle

alexeagle:
Fails on Linux too.

filipesilva:
Does this happen only on local machines? CI seems to be green.

gregmagolan:
> Does this happen only on local machines? CI seems to be green.

The firefox tests have been commented out on CI

filipesilva:
Oh that's right. On the Windows CI they seem to be running though: https://github.com/bazelbuild/rules_typescript/blob/master/.bazelci/presubmit.yml

If both firefox and chrome tests are commented out for linux/osx, does that mean that `ts_web_test` is not being tested there at all? I think those are the only browsers that are available.

gregmagolan:
Chrome is still tested in CircleCI linux at least. But yes, I think we need to get more testing up for these.

gregmagolan:
A little more information on this. rules_webtesting provides a path to the chromium executable in the test environment:

```
{
  "capabilities": {
    "browserName": "chromium",
    "goog:chromeOptions": {
      "args": [
        "--headless",
        "--use-gl=swiftshader-webgl"
      ],
      "binary": "%FILE:CHROMIUM%"
    },
    "google:wslConfig": {
      "args": [
        "--port=%WSLPORT:WSL%"
      ],
      "binary": "%FILE:CHROMEDRIVER%",
      "port": "%WSLPORT:WSL%",
      "shutdown": true,
      "status": true
    }
  },
  "environment": "local",
  "browserLabel": "@io_bazel_rules_webtesting//browsers:chromium-local",
  "webTestFiles": [
    {
      "namedFiles": {
        "CHROMEDRIVER": "io_bazel_rules_webtesting/third_party/chromedriver/chromedriver.out/chromedriver",
        "CHROMIUM": "io_bazel_rules_webtesting/third_party/chromium/chromium.out/chrome-mac/Chromium.app/Contents/MacOS/chromium",
[...]

PR #299, opened 2018-10-12
Fix firefox karma tests with support for extracting rules_webtesting archive files

gregmagolan:
Resolves https://github.com/bazelbuild/rules_typescript/issues/296

### DIFF
--- a/internal/karma/karma.conf.js
+++ b/internal/karma/karma.conf.js
@@ -26,7 +26,7 @@ try
     try {
       // Paths are relative to the 'external' folder within runfiles
       extractExe = extractExe ? path.join('external', extractExe) : extractExe;
-      archiveFile = path.join('external', archiveFile)
+      archiveFile = path.join('external', archiveFile);
       const extractedExecutablePath = path.join(process.cwd(), executablePath);
       if (!extractExe) {
         throw new Error('No EXTRACT_EXE found');
--- a/internal/karma/karma.conf.js
+++ b/internal/karma/karma.conf.js
@@ -3,7 +3,61 @@
 try
 {
   const fs = require('fs');
+  const path = require('path');
   const tmp = require('tmp');
+  const child_process = require('child_process');
+  const isWindows = /^win/i.test(process.platform);
+
+  // Helper function to find a particular namedFile
+  // within the webTestMetadata webTestFiles
+  function findNamedFile(webTestMetadata, key) {
+    let result;
+    webTestMetadata['webTestFiles'].forEach(entry => {
+      const webTestNamedFiles = entry['namedFiles'];
+      if (webTestNamedFiles && webTestNamedFiles[key]) {
+        result = webTestNamedFiles[key];
+      }
+    });
+    return result;
+  }
+
+  // Helper function to extract a browser archive
+  // and return the path to extract executable
+  function extractWebArchive(extractExe, archiveFile, executablePath) {
+    try {
+      // Paths are relative to the 'external' folder within runfiles
+      extractExe = extractExe ? path.join('external', extractExe) : extractExe;
+      archiveFile = path.join('external', archiveFile)
+      const extractedExecutablePath = path.join(process.cwd(), executablePath);
+      if (isWindows) {
+        // We can't run the EXTRACT_EXE shell script on Windows
+        if (archiveFile.endsWith('.zip')) {
+          const AdmZip = require('adm-zip');
+          const zip = new AdmZip(archiveFile);
+          // Extract to cwd
+          zip.extractAllTo('.', true);
+          // We need to chmod the executable to executable since adm-zip doesn't handle that
+          fs.chmodSync(extractedExecutablePath, 755);
+        } else {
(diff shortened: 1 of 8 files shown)


internal/karma/karma.conf.js
```diff
@@ -3,7 +3,57 @@
...
+      if (isWindows) {
+        // We can't run the EXTRACT_EXE shell script on Windows
+        const AdmZip = require('adm-zip');
+        const zip = new AdmZip(archiveFile);
+        // Extract to cwd
+        zip.extractAllTo('.', true);
+        // We need to chmod the executable to executable since adm-zip doesn't handle that
+        fs.chmodSync(extractedExecutablePath, 755);
+      } else {
+        // If we're on OSX or Linux then we can use the provided rules_webtesting
```
alexeagle: could we do it with adm-zip always? why have two different modes if one of them is broken on windows

internal/karma/karma.conf.js
```diff
@@ -3,7 +3,57 @@
...
 {
   const fs = require('fs');
+  const path = require('path');
   const tmp = require('tmp');
+  const child_process = require('child_process');
+  const isWindows = /^win/i.test(process.platform);
+
+  // Helper function to find a particular namedFile
+  // within the webTestMetadata webTestFiles
+  function findNamedFile(webTestMetadata, key) {
```
alexeagle: seems like we could remove all the helper code for the extract_exe if we always unzip ourselves?

internal/karma/karma.conf.js
gregmagolan: that was my original plan. however on Linux the archive is a tar.xz which adm-zip doesn't handle

internal/karma/karma.conf.js
gregmagolan: same as above: that was my original plan. however on Linux the archive is a tar.xz which adm-zip doesn't handle

alexeagle: what about https://www.npmjs.com/package/xz + https://www.npmjs.com/package/tar ?

I don't love adding external dependencies but also this could be a lot simpler, WDYT?

gregmagolan: Yea. Using the provided EXTRACT_EXE is not idea. I'll try those packages out.

gregmagolan: I thought about this some more and looked at the npm packages. I think we'll run into trouble supporting the various compression formats in node. On Windows we can assume it will be a `.zip` file. On linux you may get `.tar.bz2` (which is what it is currently actually and not tar.xz) or `tar.gz` or `tar.xz` in the future. If we stick to using EXTRACT_EXE then it ensures we can always extract on Linux and OSX:
```
if [[ "${BASENAME}" == *.deb ]]; then
  wrap dpkg -x "${ARCHIVE}" "${OUT_DIR}"
elif [[ "${BASENAME}" == *.tar ]]; then
  wrap tar xf "${ARCHIVE}" -C "${OUT_DIR}"
elif [[ "${BASENAME}" == *.tar.bz2 || "${BASENAME}" == *.tbz2 ]]; then
  wrap tar xjf "${ARCHIVE}" -C "${OUT_DIR}"
elif [[ "${BASENAME}" == *.tar.gz || "${BASENAME}" == *.tgz ]]; then
  wrap tar xzf "${ARCHIVE}" -C "${OUT_DIR}"
elif [[ "${BASENAME}" == *.tar.Z ]]; then
  wrap tar xZf "${ARCHIVE}" -C "${OUT_DIR}"
elif [[ "${BASENAME}" == *.zip ]]; then
  wrap unzip "${ARCHIVE}" -d "${OUT_DIR}"
else
  >&2 printf "${ARCHIVE} is not a supported file type."
  exit -1
fi
```

alexeagle: Okay I think @DrMarcII might be interested, seems like EXTRACT EXE should be fixed in windows

gregmagolan: @DrMarcII Ahhh. That's good to know. I didn't notice that. Can use EXTRACT_EXE on all platforms then 👍

gregmagolan:
@alexeagle Removed adm-zip :) Using EXTRACT_EXE on all platforms as since it is a `.bat` on Windows

internal/karma/karma.conf.js
```diff
@@ -3,7 +3,43 @@
...
+    return result;
+  }
+
+  // Helper function to extract a browser archive
+  // and return the path to extract executable
+  function extractWebArchive(extractExe, archiveFile, executablePath) {
+    try {
+      // Paths are relative to the 'external' folder within runfiles
+      extractExe = extractExe ? path.join('external', extractExe) : extractExe;
+      archiveFile = path.join('external', archiveFile)
```
alexeagle: is this going to work with the --nolegacy_external_runfiles flag?
I'll wait for that one to land before doing this one.
also missing a trailing semicolon :0

gregmagolan: thanks. I'll rebase to and let CI run on it,.

PR #306, opened 2018-10-18
Disable Chrome sandbox when unavailable

filipesilva:
Chrome on Linux uses sandboxing, which needs user namespaces to be enabled.
This is not available on all kernels and it might be turned off even if it is available.
Notable examples where user namespaces are not available include:

- In Debian it is compiled-in but disabled by default.
- The Docker daemon for Windows or OSX does not support user namespaces.

We can detect if user namespaces are supported via /proc/sys/kernel/unprivileged_userns_clone.

For more information see:
https://github.com/Googlechrome/puppeteer/issues/290
https://superuser.com/questions/1094597/enable-user-namespaces-in-debian-kernel#1122977
https://github.com/karma-runner/karma-chrome-launcher/issues/158
https://github.com/angular/angular/pull/24906

### DIFF
--- a/internal/karma/karma.conf.js
+++ b/internal/karma/karma.conf.js
@@ -41,6 +41,30 @@ try
     }
   }
 
+  // Chrome on Linux uses sandboxing, which needs user namespaces to be enabled.
+  // This is not available on all kernels and it might be turned off even if it is available.
+  // Notable examples where user namespaces are not available include:
+  // - In Debian it is compiled-in but disabled by default.
+  // - The Docker daemon for Windows or OSX does not support user namespaces.
+  // We can detect if user namespaces are supported via /proc/sys/kernel/unprivileged_userns_clone.
+  // For more information see:
+  // https://github.com/Googlechrome/puppeteer/issues/290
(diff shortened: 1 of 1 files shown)


alexeagle:
@huangw5 I think you also ran into chrome sandboxing vs. running as root - any idea if this is the right thing to do?

filipesilva:
In https://github.com/angular/angular/pull/24906 there is also a check to see if the current user is root. It could be added here.

filipesilva:
I think the problem I am trying to solve here is different because running chrome in docker on a linux host already works, but doesn't support namespaces on windows and osx hosts (breaking chrome sandboxes).

@huangw5 if I understand this correctly, running docker with `--cap-add SYS_ADMIN` means "add admin capabilities" (as per https://docs.docker.com/engine/reference/run/#runtime-privilege-and-linux-capabilities). I think that makes sense if the docker user is already root but it probably shouldn't be on when using a non-root user.

filipesilva:
Broken CI seems related to https://github.com/bazelbuild/rules_typescript/pull/312

PR #309, opened 2018-10-19
Support --nolegacy_external_runfiles bazel flag

gregmagolan:


gregmagolan:
@alexeagle Hmmmm. This works on OSX/Linux but fails on Windows. Do we make an exception for Windows?

PR #322, opened 2018-11-14
Update skydoc to latest version to use the Starlark version of http_archive

meteorcloudy:
/cc @filipesilva @alexeagle


filipesilva:
LGTM, thanks @meteorcloudy!

PR #330, opened 2018-11-27
Update README.md

lychyi:
I ran into an error using self-managed deps because a `.proto` (I think it was worker_protocol) file was not included in my `@bazel/typescript` node_modules. It turns out that it was not included in the filegroup in my root `BUILD`. Hopefully this makes the getting started process a lot smoother for folks that need self-managed deps for some reason still.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,7 @@ filegroup(
     name = "node_modules",
     srcs = glob(
         include = [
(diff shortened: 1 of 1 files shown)


lychyi:
I signed it!

lychyi:
Got it.

I couldn't come up with a more elegant reason other than "`tsc_wrapped` needs it". Hopefully, this explains it enough.

Issue #342, opened 2018-12-06
Fix Bazel incompatible changes


laurentlb:
I'd like to be able to build rules_rust with --all_incompatible_changes:
  `cd rules_typescript && bazel build --all_incompatible_changes //...`

It may not be possible to fix every issue now due to transitive dependencies, but we should try to fix the issues from this repository.

manekinekko:
@laurentlb can we close this issue now that #343 has been merged?

laurentlb:
No. In the PR description, you can see that 5 flags are disabled. The goal is to remove them.

If you want to help, look at these 3 flags:

* `--incompatible_disallow_old_style_args_add` fails due to https://github.com/bazelbuild/rules_typescript/blob/master/internal/protobufjs/ts_proto_library.bzl#L26 (args.add)
* `--incompatible_depset_union=true` fails due to https://github.com/bazelbuild/rules_typescript/blob/master/internal/common/compilation.bzl#L260
* `--incompatible_depset_is_not_iterable=false` fails due to `[f.path for f in type_blacklisted_declarations]` in https://github.com/bazelbuild/rules_typescript/blob/master/internal/common/tsconfig.bzl#L144

For more details about the flags, see https://docs.bazel.build/versions/master/skylark/backward-compatibility.html

qzmfranklin:
Any updates on this?

With version `0.22.1`, I still got errors like this:
```text
ERROR: /home/zhongming/git/LogiOcean/src/protos/BUILD:75:1: in ts_proto_library rule //src/protos:version_def.pb.ts: 
Traceback (most recent call last):
        File "/home/zhongming/git/LogiOcean/src/protos/BUILD", line 75
                ts_proto_library(name = 'version_def.pb.ts')
        File "/home/zhongming/.cache/bazel/_bazel_zhongming/2aea414f4d422ad78870bb37da8563ec/external/build_bazel_rules_typescript/internal/protobufjs/ts_proto_library.bzl", line 81, in _ts_proto_library
                _run_pbjs(ctx.actions, ctx.executable, output_..., <2 more arguments>)
        File "/home/zhongming/.cache/bazel/_bazel_zhongming/2aea414f4d422ad78870bb37da8563ec/external/build_bazel_rules_typescript/internal/protobufjs/ts_proto_library.bzl", line 30, in _run_pbjs
                args.add([f.path for f in proto_files])
[...]

PR #343, opened 2018-12-06
Update dependencies

laurentlb:
Fix a few other incompatible issues at the same time

Tested:
  `bazel test //... --all_incompatible_changes --incompatible_disable_deprecated_attr_params=false --incompatible_disallow_old_style_args_add=false --incompatible_depset_union=false --incompatible_depset_is_not_iterable=false --incompatible_no_support_tools_in_action_inputs=false`

In other words, there are still a few fixes to do.

Progress towards #342

cc @meteorcloudy @buchgr


meteorcloudy:
@alexeagle @filipesilva Can you review this change?

package.bzl
```diff
@@ -19,6 +19,7 @@ against a minimum dependent build_bazel_rules_typescript version.
 """
 
 load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+load("@bazel_tools//tools/build_defs/repo:git.bzl", "git_repository")
```
filipesilva: This load does not seem to be used.

filipesilva:
The changes look ok but it's not clear to me why the ones that are not dependency updates are needed.

Can you give me some context please?

laurentlb:
Context is the issue #342
In other words, Bazel is evolving and legacy features should be removed (https://docs.bazel.build/versions/master/skylark/backward-compatibility.html), so that rule_typescript continues working with future versions of Bazel. In many cases, dependencies have already been fixed. The PR picks up the fixes.

package.bzl
laurentlb: It was removed in the 2nd commit

filipesilva:
Ah that is indeed desirable, thanks for effecting these changes!

meteorcloudy:
@laurentlb Can you rebase this repo?
@buchgr Can you help merging this PR after rebase is done.

laurentlb:
done

alexeagle:
@buchgr This will get overwritten by the next sync from piper unless CL 224868130 lands.

PR #345, opened 2018-12-07
[WIP] Add internal profiling to tsc_wrapped worker

filipesilva:
Worker version of https://github.com/bazelbuild/rules_nodejs/pull/402.

It allows profiling TS compilation in workers, but not in single builds.


internal/tsc_wrapped/worker.ts
```diff
@@ -20,6 +21,31 @@ export function log(...args: Array<{}>) {
   console.error.apply(console, args);
 }
 
+let setup = () => Promise.resolve();
```
filipesilva: Perhaps exporting a profiling wrapper we use is a better idea.

PR #327, opened 2018-11-24
Fix devserver on windows

devversion:

PR #347, opened 2018-12-09
feat: depend on devserver binaries

mgechev:
Depend on devserver binaries instead of compiling from source. This won't require compilation of the go compiler when we start the dev server.

Once we merge this, we can release a new version and update the dependencies in the bazel package in `angular/angular`.

I went through several ideas on what would be the best way to distribute the binaries:
1. Upload the binaries to the repository and reference the `filegroup` in the `//devserver_binaries` package directly in `ts_devserver.bzl` https://github.com/bazelbuild/rules_typescript/pull/347/commits/b24f652399f4ac1ea3b60395a56afe2aa3c11418
1. (Current) Create a separate workspace with a build file which contains a `filegroup`. Create an archive for each release and distribute it in the release page https://github.com/bazelbuild/rules_typescript/pull/347/commits/0944bd7964fc6c085e32d29396f0848f8f0dfdd2


achew22:
Drive by review (sorry). Is it possible to not upload the binaries in *this* repo? If you commit the binaries in git, the repo will get big quickly for people who want to depend on `rules_typescript` at a specific commit (a situation I find myself in regularly). These binaries clock in at ~ 5mb / binary / OS / release. There are 3 OSes per release and this repo has had 74 releases since its inception. Pretending that each of those releases had a new version of the binary compiled and uploaded in the repo, that would be an additional > 1gb of binary for me to download with only 5mb of it being useful, a < 0.5% signal to noise ratio.

Have you considered using something like [Bazel toolchains](https://docs.bazel.build/versions/master/toolchains.html) to break the dependency? In the angular bazel repo, you could register a toolchain (or anyone in the dependency chain for that matter) that pulls a static binary from gstatic or downloads.google.com or even the releases page of github and leave a fallback that compiles the devserver when you're working from head. The [rules_k8s](https://github.com/bazelbuild/rules_k8s) team just finished toolchaining the `kubectl` binary, so I believe the starlark api is complete enough to support this usecase.

mgechev:
@achew22 thanks for the feedback! The binaries are not part of the repo. [Here](https://github.com/bazelbuild/rules_typescript/pull/347/files#diff-48449ecf6779b035ef4cc8783bd23133R57), you can find that I'm downloading them from a GitHub release.

achew22:
Sorry about that. I looked at [your first commit](https://github.com/bazelbuild/rules_typescript/pull/347/commits/b24f652399f4ac1ea3b60395a56afe2aa3c11418), which had the binaries checked in, and wanted to see if I could help out. The toolchain stuff is really really cool BTW. It's worth reading through the docs even just to see how the Bazel team conceptualizes toolchains. Admittedly, you do have to squint to make this look like a toolchain so if you go in a different direction, please feel free to ignore me :smile:.

internal/devserver/ts_devserver.bzl
```diff
@@ -164,7 +164,8 @@ ts_devserver = rule(
         ),
         "_requirejs_script": attr.label(allow_single_file = True, default = Label("@build_bazel_rules_typescript_devserver_deps//node_modules/requirejs:require.js")),
         "_devserver": attr.label(
-            default = Label("//devserver"),
+            # In development use Lavel("//devserver")
```
devversion: This looks like a typo. `Label`?

Also even though this PR builds Windows binaries, Windows devserver doesn't work 😄we need to move #327 forward to get it work. Nevertheless, makes sense to build the binaries already. Very cool!

package.bzl
```diff
@@ -51,6 +51,13 @@ def rules_typescript_dependencies():
         sha256 = "a13af63e928c34eff428d47d31bafeec4e38ee9b6940e70bf2c9cd47184c5c16",
     )
 
+    _maybe(
+        http_archive,
+        name = "go_devserver",
+        urls = ["https://github.com/bazelbuild/rules_typescript/releases/download/%s/devserver_binaries.zip" % VERSION],
```
alexeagle: I don't think we should do it this way. It means the user downloads the rules_typescript as a zip from github, then we later download another zip from github.
Instead I think we want the entire rules_typescript to be distributed as a zip file that we publish, which contains the current WORKSPACE and BUILD/.bzl files

mgechev: I'll update the dependency to point to `//devserver_binaries`.

internal/devserver/ts_devserver.bzl
```diff
@@ -164,7 +164,7 @@ ts_devserver = rule(
         ),
         "_requirejs_script": attr.label(allow_single_file = True, default = Label("@build_bazel_rules_typescript_devserver_deps//node_modules/requirejs:require.js")),
         "_devserver": attr.label(
-            default = Label("//devserver"),
+            default = Label("//devserver:server"),
```
mgechev: @alexeagle let me know if you have better suggestions for this dependency.

BUILD.bazel
```diff
@@ -70,3 +71,31 @@ nodejs_binary(
...
+    name = "release",
+    srcs = [
+      "BUILD.bazel",
+      "AUTHORS",
+      "defs.bzl",
+      "LICENSE",
+      "on-version.js",
+      "package.bzl",
+      "WORKSPACE",
+      "yarn.lock",
```
alexeagle: No use publishing lockfile right?

BUILD.bazel
```diff
@@ -70,3 +71,31 @@ nodejs_binary(
...
 )
+
+pkg_tar(
+    name = "release",
+    srcs = [
+      "BUILD.bazel",
+      "AUTHORS",
+      "defs.bzl",
+      "LICENSE",
+      "on-version.js",
```
alexeagle: Is this used?

PR #334, opened 2018-12-04
docs: add community documentation

mgechev:
This PR introduces:

1. Issue templates for bug reports and feature requests
1. Pull request template
1. Code of conduct, which reuses the content from angular/angular
1. Guidelines for contributing to the project
1. A development document which explains how to get started with fixing
issues/implementing new rules. There's more work here, this is just the
initial draft
1. Update of the readme to introduce project scope

PR #348, opened 2018-12-10
WIP: Support for compiling tsc_wrapped with typescript 3.1

gregmagolan:


internal/tsc_wrapped/cache.ts
```diff
@@ -127,7 +127,7 @@ class Cache<T> {
   }
 }
 
-interface SourceFileEntry {
+export interface SourceFileEntry {
```
gregmagolan: interface needs exporting as its used as a parameter in an exported class

internal/tsc_wrapped/compiler_host.ts
```diff
@@ -440,7 +440,7 @@ export class CompilerHost implements ts.CompilerHost, tsickle.TsickleHost {
   writeFile(
       fileName: string, content: string, writeByteOrderMark: boolean,
       onError: ((message: string) => void) | undefined,
-      sourceFiles: ReadonlyArray<ts.SourceFile>): void {
+      sourceFiles: ReadonlyArray<ts.SourceFile> | undefined): void {
```
gregmagolan: Fixes 
```
external/build_bazel_rules_typescript/internal/tsc_wrapped/compiler_host.ts(446,3): error TS2416: Property 'writeFile' in type 'CompilerHost' is not assignable to the same property in base type 'CompilerHost'.
  Type '(fileName: string, content: string, writeByteOrderMark: boolean, onError: ((message: string) => void) | undefined, sourceFiles: ReadonlyArray<SourceFile>) => void' is not assignable to type 'WriteFileCallback'.
    Types of parameters 'sourceFiles' and 'sourceFiles' are incompatible.
      Type 'ReadonlyArray<SourceFile> | undefined' is not assignable to type 'ReadonlyArray<SourceFile>'.
        Type 'undefined' is not assignable to type 'ReadonlyArray<SourceFile>'.
```
My guess is that in ts 3.1 if a preceding parameter can be undefined (onError in this case) then subsequent parameters are assumed to be allowed to be undefined?

`writeFile` is defined as `type WriteFileCallback = (fileName: string, data: string, writeByteOrderMark: boolean, onError: ((message: string) => void) | undefined, sourceFiles?: ReadonlyArray<SourceFile>) => void;` in typescript 3.1 so its not strictly a call signature mismatch but maybe a change in how call signatures are interpreted?

internal/tsetse/language_service_plugin.ts
```diff
@@ -13,22 +12,28 @@ function init() {
...
+      const program = oldService.getProgram();
 
-      // Add disabledRules to tsconfig to disable specific rules
-      // "plugins": [
-      //   {"name": "...", "disabledRules": ["equals-nan"]}
-      // ]
-      registerRules(checker, info.config.disabledRules || []);
+      // Signature of `getProgram` is `getProgram(): Program | undefined;` in ts 3.1
+      // so we must check if the return value is valid to compile with ts 3.1.
+      if (program) {
```
gregmagolan: I suspect an undefined `program` would cascade and things would fail anyway so maybe the best thing to do is to throw in this case?

internal/tsetse/rules/ban_promise_as_condition_rule.ts
```diff
@@ -95,10 +95,15 @@ function addFailureIfThenableCallExpression(
 
   const typeChecker = checker.typeChecker;
   const signature = typeChecker.getResolvedSignature(callExpression);
-  const returnType = typeChecker.getReturnTypeOfSignature(signature);
 
-  if (isNonFalsyThenableType(typeChecker, callExpression, returnType)) {
-    checker.addFailureAtNode(callExpression, errorMessage);
+  // Return value of getResolvedSignature is `Signature | undefined` in ts 3.1
+  // so we must check if the return value is valid to compile with ts 3.1.
+  if (signature) {
```
gregmagolan: Same as above. Throw here if signature is undefined? Don't know enough about the typeChecker to know which is better

DEVELOPING.md
```diff
@@ -0,0 +1,66 @@
+# For Developers
+
+We strongly encourage you to review the project's scope described in the `README.md` file before working on new features. For large changes, consider writing a design document using [this template](https://docs.google.com/document/d/1U6GlAaPF3tl1wvBNXebW1mD3oHZxV5h9TH-bR0p8Uz8/edit?usp=sharing).
+
+## Testing changing downstream
+
+By defaults, downstream projects use both an `http_archive` of `build_bazel_rules_typescript` and the released `@bazel/typescript` and `@bazel/karma` npm packages. `postinstall` steps in these npm packages check that the version of the build_bazel_rules_typescript is compatible with the version of the npm package(s).
```
gregmagolan: "By default,"

internal/tsetse/language_service_plugin.ts
filipesilva: Yeah throwing seems the correct behaviour here.

DEVELOPING.md
```diff
@@ -0,0 +1,67 @@
...
+to
+
+```python
+"compiler": attr.label(
+    default = Label("@build_bazel_rules_typescript//internal:tsc_wrapped_bin"),
+```
+
+The correct defaults to use so that you are not depending on the npm package downstream are in `/internal/defaults.bzl`. Note, your downstream
+workspace will also need the correct `@npm` dependencies available to build these targets (see `internal/e2e/package_typescript_3.1_no_npm/package.json`).
+In the case of the `angular` workspace, some `@npm` dependencies in this repository will also need to be changed to `@ngdeps` since `angular` does not have
```
filipesilva: At first it's somewhat surprising that this is needed, but it makes sense. 

The npm package installed downstream was using the `typescript` npm peerDependency. So we need to tell the rule to do the same.

I wonder if there's a more ergonomic way of doing this, like saving `"npm"` in a variable then using something like `"@%s//typescript" % npmLabel`,". I'm not sure of who would be doing this besides us though.

DEVELOPING.md
```diff
@@ -0,0 +1,67 @@
+# For Developers
+
+We strongly encourage you to review the project's scope described in the `README.md` file before working on new features. For large changes, consider writing a design document using [this template](https://docs.google.com/document/d/1U6GlAaPF3tl1wvBNXebW1mD3oHZxV5h9TH-bR0p8Uz8/edit?usp=sharing).
```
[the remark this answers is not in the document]
gregmagolan: Removed that line since its from #334 and will make it in with that PR

internal/tsetse/language_service_plugin.ts
gregmagolan: 👍

internal/tsetse/rules/ban_promise_as_condition_rule.ts
gregmagolan: Throwing here as well

gregmagolan:
Note to caretaker: `internal/e2e/ts_auto_deps/tsconfig.json.oss` renamed to `internal/e2e/ts_auto_deps/tsconfig.json`. I think the sync needs to be configured to prevent the automatic rename.

internal/tsc_wrapped/cache.ts
gregmagolan: true for ts 2.7 at least. this is no longer an error as of ts 2.9. didn't check 2.8.

PR #349, opened 2018-12-11
Use buildifier package from npm

alexeagle:


alexeagle:
I'll let Wassim own this one...

PR #352, opened 2018-12-12
Update @bazel/typescript and @bazel/karma version check to use semver…

gregmagolan:
… to match only major & minor version against bazel package version

Also, updates the version in package.bzl and package.json back to 0.22.0 after a regression to 0.21.0 via sync.


gregmagolan:
NOTE: removed to yarn.lock files that are not used and not needed

PR #355, opened 2018-12-14
Do not depend on the go rules

### LINKED ISSUE #353: Do not depend on the Go bazel rules
Developers using rules_typescript should not depend on the Go bazel rules. This will require creating a different file layout inside of the release package. In a `devserver_binaries` we should expose a BUILD file which points to the produced platform binaries but does not depend on the Go bazel rules.

mgechev:
## PR Checklist

- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [x] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes

- The users of `rules_typescript` need the Go rules although they don't use them
- `http_archive` fails when we produce a package with `tgz` extension
- Initializing the workspace in an Angular CLI project generated with the Bazel schematics fails without `internal/devserver/(yarn.lock|package.json)`. Adding the files might not be required, I'll verify before removing the WIP.

Issue Number: Fix #353

## What is the new behavior?

- We no longer depend on the Go rules
- We package the release as a `tar` file
- Includes `internal/devserver/(yarn.lock|package.json)` in the package. Before removing WIP, I'll double check if this is required.

- [ ] Yes
- [x] No


mgechev:
@alexeagle I made the test pass locally but this pattern `filename` and `filename-package` emerges more and more often. I'll think of a way to approach in a more idiomatic way so we can discuss this on Monday.

alexeagle:
I filed https://github.com/bazelbuild/rules_webtesting/issues/319 for the remaining transitive dependency on rules_go

README.md
```diff
@@ -77,11 +77,6 @@ yarn_install(
   yarn_lock = "//:yarn.lock",
 )
 
-# Setup Go toolchain
```
mgechev: Only webtesting depends on Go, so this should be valid

alexeagle:
Greg included this in the c3cec89bc0e90b0d6da1c96fa16ba095503ea82e

PR #356, opened 2018-12-14
Make powershell.exe available at build time

meteorcloudy:
After flipping `--experimental_strict_action_env`, we need to pass `--action_env=PATH` to make sure some specific tools are available on Windows.
Fixing https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/4#a868eb9b-dc52-4ce1-8adb-0da82e8a407b

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -40,8 +40,12 @@ platforms:
     run_targets:
     - "//:gazelle"
     - "@nodejs//:yarn"
+    build_flags:
+    - "--action_env=PATH"
     build_targets:
     - "..."
     - "@disable_tsetse_for_external_test//..."
+    test_flags:
+    - "--test_env=PATH"
     test_targets:
     - "..."


meteorcloudy:
Another though, maybe it's better to add `--action_env` in the bazelrc file directly?

buchgr:
Hmm. Is powershell always available in the same place on windows? Would something like `--action_env=PATH=<C:\path\to\powershell.exe` work? I think passing in the whole path is typically not recommended, because it's not compatible with remote caching / execution.

meteorcloudy:
On my Windows workstation:
```
pcloudy@PCLOUDY0-W C:\tools\msys64\home\pcloudy
> where powershell.exe
C:\Windows\System32\WindowsPowerShell\v1.0\powershell.exe
```

Looks like it depends on the powershell version, so it's not always the same.
Also, currently we set PATH as msys bin path when `--experimental_strict_action_env`, this make bash tools available, will `--action_env=PATH=<C:\path\to\powershell.exe>` override it?

meteorcloudy:
Yes, I think that would be very useful.
But for now, we still need this change to fix the Windows build, right?

alexeagle:
@meteorcloudy could you send this as a CL to third_party/bazel_rules/rules_typescript? see the README.google.md in that directory if you'd like to use copybara to import this PR

filipesilva:
Powershell is not guaranteed to be on any given machine though. The windows nanoserver docker image does not contain it, for instance. It also does not support msys2 anyway, thus also does not support Bazel, but it's an example.

Is powershell a requirement now for Bazel?

filipesilva:
@alexeagle the error in https://buildkite.com/bazel/bazel-at-head-plus-disabled/builds/4#a868eb9b-dc52-4ce1-8adb-0da82e8a407b doesn't seem related to the yarn path issue (https://github.com/angular/angular/issues/27514). This is just a missing binary (powershell), whereas the other one was the env path changing on every **run..**

But I agree that perhaps the `--action_env=PATH` flag can help work around it.

meteorcloudy:
@filipesilva powershell is not a requirement for Bazel, but it should be a native tool on Windows, which means users usually don't have to install it. But I agree we should avoid it if possible, so that we don't have to use `--action_env=PATH` to work around.

meteorcloudy:
@alexeagle Yes, they'll need to add `--action_env=PATH` for building on Windows. So, do you think it's better to add `--action_env=PATH` to the [.bazelrc](https://github.com/bazelbuild/rules_typescript/blob/master/.bazelrc) file? If so, users don't have to do anything, but the downside is this will also affect other platforms.

filipesilva:
I really want to echo what @achew22 asked: 
> Why is this change necessary on a per project basis instead of being global? Can you also help me understand why isn't powershell treated like bash and made available without any flags? Is a change like this coming for Mac/Linux soon?

It's a bit surprising, and onerous, that users need to effect this change to build on windows but shouldn't add it to build on other platforms. 

Can't a non-powershell script be used instead of requiring this change downstream on all projects?

alexeagle:
merged in d53748ce14074543b265803c1b7a2eaea4f03271

ocombe:
I just got the following error on windows: `FIND: Parameter format not correct`, based on https://serverfault.com/questions/758267/powershell-pipe-into-find-exe-command it seems that powershell evaluated anything that is in double quotes, but the find command requires double quotes, meaning that we should escape them to avoid that powershell evaluates the expression.
Like @filipesilva said, we really should not use powershell, it's going to be a whole lot of problems like this one...

meteorcloudy:
The powershell usage comes from https://github.com/bazelbuild/rules_webtesting/blob/c46c3f94aaee0d0a2d5bb2150d46b64bc798f0e2/web/internal/extract.bat

/cc @DrMarcII, is there any other way to unzip a file on Windows?

PR #363, opened 2018-12-19
Add karma_web_test and karma_web_test_suite

gregmagolan:
Adds karma_web_test and karma_web_test_suite which accepts a user base karma configuration in `config_file` attribute. ts_web_test (a generic web_test rule) uses karma_web_test code under the hood but does not accept a karma configuration. Updated `examples/testing:testing` to also show karm_web_test usage.

Also some general cleanup in the karma rule.

Prerequisite for https://github.com/angular/angular/pull/27721.


gregmagolan:
Also added environment variables support base karma configuration so you can do `outputFile: "${TEST_UNDECLARED_OUTPUTS_DIR}/karma-result.json",` which is pretty cool.

And added `configuration_env_vars` as well so I could make the following work in angular/angular:

`$ yarn bazel test //:test_web_all --define=TS_WEB_TEST_MODE=SL_REQUIRED`

karma-js.conf.js:
```
  switch (process.env.TS_WEB_TEST_MODE) {
    case 'SL_REQUIRED':
      conf.browsers = browserProvidersConf.sauceAliases.CI_REQUIRED;
      break;
    case 'SL_OPTIONAL':
      conf.browsers = browserProvidersConf.sauceAliases.CI_OPTIONAL;
      break;
    case 'BS_REQUIRED':
      conf.browsers = browserProvidersConf.browserstackAliases.CI_REQUIRED;
      break;
    case 'BS_OPTIONAL':
      conf.browsers = browserProvidersConf.browserstackAliases.CI_OPTIONAL;
      break;
    default:
      conf.browsers = [process.env['DISPLAY'] ? 'Chrome' : 'ChromeHeadless'];
  }
```

Test target of SauceLabs, BrowserStack or local chrome controlled by `-define=TS_WEB_TEST_MODE=`

gregmagolan:
That's a good question. The plugin would have to be able to be able to override configuration values such as `browsers`, `customLaunchers`,`sauceLabs`, `browserStack` etc... (see https://github.com/angular/angular/pull/27721/files#diff-c4667ceccd7e91092048e752850fab9c) which would make it functionally the same as a base configuration except harder to setup for users.

defs.bzl
```diff
@@ -23,9 +23,14 @@ load("//internal:ts_config.bzl", _ts_config = "ts_config")
 load("//internal/devserver:ts_devserver.bzl", _ts_devserver = "ts_devserver_macro")
 load(
     "//internal/karma:ts_web_test.bzl",
-    _ts_web_test = "ts_web_test_macro",
+    _ts_web_test = "ts_web_test",
     _ts_web_test_suite = "ts_web_test_suite",
 )
+load(
+    "//internal/karma:karma_web_test.bzl",
```
alexeagle: not sure we should put this in the public API yet.

I think generally we should have some process for maturing new rules before they go in the public API

examples/testing/karma.conf.js
```diff
@@ -0,0 +1,12 @@
+module.exports = function(config) {
+  config.set({
+    plugins: ['karma-json-result-reporter'],
+    reporters: ['dots', 'progress', 'json-result'],
+    logLevel: config.LOG_DEBUG,
+    colors: false,
+    jsonResultReporter: {
+      outputFile: "${TEST_UNDECLARED_OUTPUTS_DIR}/karma-result.json",
```
kyliau: Is there a reason for the deferred variable substitution? If so, can you please add an explanation here?
Since this file is executed under Bazel (it is require'd in the same context as the base config file), it should have access to the environment variables and they should be used directly.

gregmagolan: Right. That's not necessary. Thanks!

PR #365, opened 2018-12-21
Support for ngtsc as a ts_library plugin

alexeagle:


alexeagle:
landing this in other PRs

PR #371, opened 2019-01-01
fix(karma): do not show error message if unnamed amd module is loaded

devversion:
Currently unnamed AMD modules need to be manually configured through a `require.config` call. This then configures Require.JS to load the unnamed AMD module dynamically as a "static file". Meaning that the unnamed AMD module needs to be specified within the `ts_web_test_suite` `static_files` attribute.

All static files will be proxied by default and can be loaded through `${location}/base/{workspace}/{path}` and therefore it's common to load static files that way.

**The problem** here is that the "karma-requirejs" plugin by default shows an error message if something is loaded through a karma proxy. We need to disable this warning for such known & proxied requests so that the error output is clean and does not cause confusion what "no timestamp for XXX" means.


devversion:
@mgechev I've rebased this one again (now that master is green). Let's see if this one turns green as well now.

PR #374, opened 2019-01-04
Generate npm packages that use rootPath to install sibling bazel workspaces

gregmagolan:
Introduces new build_bazel_rules_karma repository that is installed by the @bazel/karma npm package.


gregmagolan:
@alexeagle This is also difficult to downstream test without a tagged release since there are no published npm packages to fetch downstream

gregmagolan:
//cc @mgechev I reduced the release pkg_tar to use the output of the npm_package so two parallel structures don't need to be maintained. The trouble is that the `npm_package` directory is included in the archive and `strip_prefix` does not work in this case. Haven't found a solution for this yet.

BUILD.bazel
```diff
@@ -18,10 +18,19 @@
...
+# Generated `@bazel/typescript` npm package should contain a trimmed BUILD file.
+
 load("@bazel_gazelle//:def.bzl", "gazelle")
-load("@bazel_tools//tools/build_defs/pkg:pkg.bzl", "pkg_tar")
-load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary")
+
+# END-INTERNAL
+load("@build_bazel_rules_nodejs//:defs.bzl", "nodejs_binary", "npm_package")
+
+# BEGIN-INTERNAL
```
alexeagle: external users depend on `protobufjs_bootstrap_scripts` which is now fenced

BUILD.bazel
```diff
@@ -55,39 +64,68 @@ js_library(
     visibility = ["//visibility:public"],
 )
 
+npm_package(
+    name = "npm_package",
+    srcs = [
+        "AUTHORS",
+        "BUILD.bazel",
+        "CODE_OF_CONDUCT.md",
```
alexeagle: let's minify what we publish, I don't think we need things like CONTRIBUTING.md - if you want to contribute you'll come to our repo, not look through the distribution

BUILD.bazel
```diff
@@ -55,39 +64,68 @@ js_library(
...
+    deps = [
+        # TODO(mgechev): add devserver binaries back in once they work (see above)
+        # "//devserver:devserver-darwin",
+        # "//devserver:devserver-linux",
+        # "//devserver:devserver-windows",
+        "//internal:tsc_wrapped",
+    ],
+)
+
+# Produces the release we publish to GitHub. Users download this starlark package
```
alexeagle: are we still going to publish a release to github? If you're not using the `@bazel/typescript` package I'd rather you just fetch our whole repo (and build things from source)

DEVELOPING.md
```diff
@@ -60,13 +60,13 @@ rule is that minors are breaking changes and patches are new features).
 1. May be necessary if Go code has changed though probably it was already necessary to run this to keep CI green: `bazel run :gazelle`
 1. If we depend on a newer rules_nodejs, update the `check_rules_nodejs_version` in `ts_repositories.bzl`
 1. `git commit -a -m 'Update docs for release'`
-1. `npm config set tag-version-prefix ''`
-1. `npm version minor -m 'rel: %s'` (replace `minor` with `patch` if no breaking changes)
-1. Build npm packages and publish them: `bazel run //internal:npm_package.publish && bazel run //internal/karma:npm_package.publish`
-1. `bazel build :release`
+1. Update `VERSION` in `version.bzl` (increment `patch` if no breaking changes, otherwise increment `minor`)
```
alexeagle: let's do a separate commit on the release instructions after we successfully do a release and have refined the process

ts_auto_deps/BUILD.bazel
```diff
@@ -16,3 +18,16 @@ go_binary(
     embed = [":go_default_library"],
     visibility = ["//visibility:public"],
 )
+
+filegroup(
```
alexeagle: don't include any .go files in the distro. see recent commit: we now have pre-built binary for ts_auto_deps

gregmagolan: 👍

DEVELOPING.md
gregmagolan: 👍

BUILD.bazel
gregmagolan: Ideally we can archive up the generated npm package and publish two archives: one for rules_typescript and one for rules_karma. The trouble is the `pkg_tar` rules adds the `npm_package` folder to the archive.

Alternately, users can http_archive the repo for rules_typescript and http_archive the repo with a strip-prefix of `internal/karma` for the rules_karma workspace.

BUILD.bazel
gregmagolan: 👍

PR #376, opened 2019-01-08
Update yarn lock file after change to typescript 3.1

gregmagolan:


gregmagolan:
Buildkite failure on Windows is unrelated. It was introduced with https://github.com/bazelbuild/rules_typescript/pull/356. And the CircleCI test failure is due to memory exhaustion on the default CircleCI container. I've re-run it a number of times with no luck on the default 4MB container.

gregmagolan:
Update: all green after rebase to master which was switched to xlarge resource_class for test job. Karma tests on Windows are green now as well.

PR #378, opened 2019-01-09
Fix for Bazel incompatible changes

### LINKED ISSUE #342: Fix Bazel incompatible changes

laurentlb:
- Update to the new args API
- Update the dependencies
- Avoid implicit depset flattening
- Avoid depset concatenation

Fixes #342

Tested:
  `bazel-0.21 test //... --all_incompatible_changes --incompatible_disable_deprecated_attr_params=false`

There's one remaining issue, but it's in the protobuf dependency.

### DIFF
--- a/internal/build_defs.bzl
+++ b/internal/build_defs.bzl
@@ -103,7 +103,7 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file, node_opts, description 
     # These deps are identified by the NodeModuleInfo provider.
     for d in ctx.attr.deps:
         if NodeModuleInfo in d:
-            action_inputs.extend(_filter_ts_inputs(d.files))
+            action_inputs.extend(_filter_ts_inputs(d.files.to_list()))
 
     if ctx.file.tsconfig:
         action_inputs.append(ctx.file.tsconfig)
--- a/internal/common/compilation.bzl
+++ b/internal/common/compilation.bzl
@@ -257,7 +257,7 @@ def compile_ts(
         tsickle_externs = [ctx.actions.declare_file(ctx.label.name + ".externs.js")]
 
     dep_declarations = _collect_dep_declarations(ctx, deps)
-    input_declarations = dep_declarations.transitive + src_declarations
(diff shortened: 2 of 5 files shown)


laurentlb:
CI error is `error An unexpected error occurred: "ENOMEM: not enough memory`. Can we rerun the test? (I suspect it's unrelated)

filipesilva:
I think the circleci job is being killed because of a a high memory ceiling. Looking at it in https://github.com/bazelbuild/rules_typescript/pull/380.

filipesilva:
Heya, can you rebase on master? The memory problem has a fix there now.

laurentlb:
Tests are now passing, thanks.

internal/build_defs.bzl
```diff
@@ -103,7 +103,7 @@ def _compile_action(ctx, inputs, outputs, tsconfig_file, node_opts, description
     # These deps are identified by the NodeModuleInfo provider.
     for d in ctx.attr.deps:
         if NodeModuleInfo in d:
-            action_inputs.extend(_filter_ts_inputs(d.files))
+            action_inputs.extend(_filter_ts_inputs(d.files.to_list()))
```
manekinekko: `to_list()` has some performance issues and should be avoided ([see ref](https://docs.bazel.build/versions/master/skylark/performance.html#avoid-calling-depsetto_list)).

alexeagle: I think this is subject to the performance problem either way. The question is whether `to_list` is done implicitly by passing the depset to an API that expects a list, or done explicitly. I think the bazel team is moving us to the latter.

laurentlb: Yes, it was just to make the behavior more obvious. Bazel implicitly flattens depsets in some places (which is a performance problem). We want to forbid that. The simplest way forward is to make all flattening explicit in existing files.

PR #385, opened 2019-01-18
fix: devserver not working on windows

devversion:


devserver/devserver/devserver.go
```diff
@@ -175,34 +169,40 @@ func CreateFileHandler(servingPath, manifest string, pkgs []string, base string)
...
 
-			// Continue searching if the runfile couldn't be found for the request filed.
-			if _, statErr := os.Stat(realFilePath); err != nil || statErr != nil {
+			// Continue searching if the runfile couldn't be found for the requested file.
+			if err != nil {
 				continue
 			}
 		}
 
+		stat, err := os.Stat(realFilePath)
```
mgechev: Internally `Runfile` checks for the existence of the file. I think this check is redundant.

mgechev: If `err` gotten from `runfiles.Runfile` above is not `nil`, `realFilePath` will always have an invalid value.

devversion: Yeah. We just continue looping if that's the case. do you want me to change something or do I miss something?

PR #388, opened 2019-01-28
fix: error messages in compiler host

alan-agius4:


kyliau:
Merged in 9fcff2bfcc4872f28833ec772ef082e1dd2a9beb

PR #395, opened 2019-01-31
Cleanup of e2e tests

gregmagolan:
Cleans up e2e tests.

Tests that depend on generated npm packages are run with `yarn e2e-npm_packages` which calls the shell script `./internal/e2e/npm_packages/test.sh`. This script generates the `package.json` file for each test using absolute paths for the generated `@bazel/typescript` and `@bazel/karma` npm packages before running the tests.


internal/e2e/npm_packages/test.sh
```diff
@@ -0,0 +1,53 @@
...
+BAZEL=$(pwd)/node_modules/.bin/bazel
+if [[ ! -f $BAZEL ]] ; then
+  echo "Bazel not found under $BAZEL"
+  exit 1
+fi
+BAZEL_BIN=$($BAZEL info bazel-bin)
+BAZEL_TYPESCRIPT_NPM_PACKAGE=$BAZEL_BIN/internal/npm_package
+BAZEL_KARMA_NPM_PACKAGE=$BAZEL_BIN/internal/karma/npm_package
+echo "@bazel/typescript: $BAZEL_TYPESCRIPT_NPM_PACKAGE"
+echo "@bazel/karam: $BAZEL_KARMA_NPM_PACKAGE"
```
alexeagle: karma misspelled

internal/e2e/npm_packages/test.sh
```diff
@@ -0,0 +1,53 @@
...
+    if [[ ! -f "package-template.json" ]] ; then
+      echo "No package-template.json file found in $testDir"
+      exit 1
+    fi
+    # Generate package.json subsituting variables
+    # BAZEL_TYPESCRIPT_NPM_PACKAGE and BAZEL_KARMA_NPM_PACKAGE
+    ESCAPED_TYPESCRIPT=$(echo $BAZEL_TYPESCRIPT_NPM_PACKAGE | sed 's/\//\\\//g')
+    ESCAPED_KARMA=$(echo $BAZEL_KARMA_NPM_PACKAGE | sed 's/\//\\\//g')
+    sed -e "s/\${BAZEL_TYPESCRIPT_NPM_PACKAGE}/$ESCAPED_TYPESCRIPT/" -e "s/\${BAZEL_KARMA_NPM_PACKAGE}/$ESCAPED_KARMA/" package-template.json > package.json
+    if [[ $# -ge 1 && $1 = "--update-lock-files" ]] ; then
```
alexeagle: maybe print this option somewhere so it's easier for us to remember how to do it?

internal/e2e/npm_packages/test.sh
```diff
@@ -0,0 +1,53 @@
...
+      # Update yarn.lock files
+      yarn install
```
alexeagle: don't we always need to run an install?
we might want to use the option (is it frozen-lockfile?) that complains if the lockfile appears out-of-date

package.json
```diff
@@ -38,23 +38,16 @@
         "which": "~1.0.5"
     },
     "scripts": {
-        "pree2e": "webdriver-manager update $CHROMEDRIVER_VERSION_ARG && bazel build examples/app:e2e && bazel build examples/protocol_buffers:e2e",
-        "e2e": "yarn e2e-bazel-external && yarn e2e-examples-app-devserver && yarn e2e-examples-app-prodserver && yarn e2e-examples-protobuf-devserver && yarn e2e-examples-protobuf-prodserver && yarn e2e-package_karma && yarn e2e-package_typescript && yarn e2e-package_typescript_karma && yarn e2e-ts_auto_deps",
+        "pree2e": "webdriver-manager update $CHROMEDRIVER_VERSION_ARG && bazel build examples/app:e2e && bazel build examples/protocol_buffers:e2e && bazel build //internal:npm_package && bazel build //internal/karma:npm_package",
```
alexeagle: why do we have four separate bazel build commands instead of one build command with four targets? the latter could parallelize better

internal/e2e/npm_packages/test.sh
gregmagolan: Using --frozen-lockfile would be an issue here since the absolute paths to the npm packages injected into package.json will be different on each machine

internal/e2e/npm_packages/test.sh
gregmagolan: 👍

internal/e2e/npm_packages/test.sh
gregmagolan: 👍

package.json
gregmagolan: 👍

PR #402, opened 2019-02-04
Add missing fencing

gregmagolan:
Fixes build after missing code fencing that got missed after syncing #374


internal/BUILD.bazel
```diff
@@ -14,6 +14,9 @@
 
 # gazelle:exclude worker_protocol.proto
 
+# BEGIN-DEV-ONLY
```
alexeagle: We don't want any content from this file, right? How about we just generate an empty marker file in the build?

gregmagolan:
Closed by 12b12d758b6ce3aedcd149e083a1e00c4df87d37

PR #416, opened 2019-02-12
Generate npm_packages in internal/e2e/npm_packages/test.sh for smoother local testing

gregmagolan:
This is so you don't have to call `yarn pree2e` before calling `yarn e2e-npm_packages` when running e2e tests locally


kyliau:
sync cl: https://critique.corp.google.com/#review/233541758

PR #418, opened 2019-02-12
Pin rules_webtesting browser versions for CI and local testing

gregmagolan:
The default chrome browser provided by rules_webtesting browser_repositories() is broken on OSX. This doesn't show up in CI since we're only running the e2e tests on CircleCI but it does break for me locally when running the e2e tests on OSX. This PR resolves that.

//cc @alexeagle 

### DIFF
--- a/WORKSPACE
+++ b/WORKSPACE
@@ -93,82 +93,10 @@ load("@io_bazel_rules_webtesting//web:repositories.bzl", "web_test_repositories"
 
 web_test_repositories()
 
-# Load pinned browser versions for rules_webtesting.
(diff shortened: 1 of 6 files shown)


alexeagle:
mailed as cl/233760266 to @kyliau

gregmagolan:
Updating to the latest rules_typescript is blocked on angular/angular as well by the es2015 change (see https://github.com/angular/angular/pull/28625). In the meantime, I'll pin the same browsers in angular/angular with another PR since we can't update rules_typescript yet.

Issue #377, opened 2019-01-08
tsickle warnings are treated as errors


JaredNeil:
# 🐞 bug report

### Description

Warnings generated by `tsickle` are treated as errors, causing the build to fail.

## 🔬 Minimal Reproduction

1. Check out this branch: [https://github.com/JaredNeil/angular-bazel-example/tree/jaredneil-tsickle-warning](https://github.com/JaredNeil/angular-bazel-example/tree/jaredneil-tsickle-warning)
2. Run `bazel build //src/reducers:reducers`
3. Notice only warnings are printed, but the build fails.

I expected to see only warnings printed, and the build to pass.

## 🔥 Exception or Error

<pre><code>ERROR: /home/jaredneil/code/angular-bazel-example/src/reducers/BUILD.bazel:5:1: Compiling TypeScript (devmode) //src/reducers:reducers failed (Exit 1)
src/reducers/reducers.ts:38:1 - warning TS0: @namespace annotations are redundant with TypeScript equivalents

38 /** @namespace */
   ~~~~~~~~~~~~~~~~~

Target //src/reducers:reducers failed to build
</code></pre>


**Rules version (SHA):**

`2963b55370b21d545d0ac0f30fca9dc74a0f5538`

PR #419, opened 2019-02-14
Don't treat tsickle warnings as errors

JaredNeil:
Not sure if you want to merge this or not, but figured I'd put it here anyway.
Happy to add a new test if you want me to before merging.

## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [x] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_typescript/issues/377

## What is the new behavior?

```
$ bazel build //src/reducers:reducers
INFO: From Compiling TypeScript (devmode) //src/reducers:reducers:
src/reducers/reducers.ts:38:1 - warning TS0: @namespace annotations are redundant with TypeScript equivalents

38 /** @namespace */
   ~~~~~~~~~~~~~~~~~

Target //src/reducers:reducers up-to-date:
  dist/bin/src/reducers/reducers.d.ts
INFO: Build completed successfully, 3 total actions
```

- [ ] Yes
- [x] No

### DIFF
--- a/internal/tsc_wrapped/tsc_wrapped.ts
+++ b/internal/tsc_wrapped/tsc_wrapped.ts
@@ -241,9 +241,13 @@ function runFromOptions(
   } else {
     diagnostics = emitWithTypescript(program, compilationTargets);
   }
-
-  if (diagnostics.length > 0) {
-    console.error(bazelDiagnostics.format(bazelOpts.target, diagnostics));
+  const warnings = diagnostics.filter(d => d.category == ts.DiagnosticCategory.Warning);
+  if (warnings.length > 0) {
+    console.warn(bazelDiagnostics.format(bazelOpts.target, warnings));
+  }
+  const errors = diagnostics.filter(d => d.category == ts.DiagnosticCategory.Error);
+  if (errors.length > 0) {
+    console.error(bazelDiagnostics.format(bazelOpts.target, errors));
     debug('compilation failed at', new Error().stack!);
     return false;
   }


alexeagle:
hmm @evmar just made some changes around this same thing, so that we always treat these as errors in google3
can you explain more why you need this? It's easier for our cognitive load if google3 and external work the same.

evmar:
These tsickle warnings happen when code you've written does something that Closure compiler would interpret and we think you shouldn't have included it.  But for non-Google users who aren't necessarily targeting Closure compiler, it's not clear if we should fail their builds because of it.  On the other hand, printing warnings is kind of annoying as well.

I think maybe we should just swallow+hide these warnings entirely.  I wonder if maybe tsickle should return its warnings via some side channel so that it doesn't get mixed in with compiler warnings.

The other thing we talked about is that tsickle really shouldn't be in the business of complaining about this.  It sorta belongs in tsetse instead.  But the problem with that is that it would mean duplicating a bunch of tsickle logic into tsetse (to identify the problematic patterns).

What we recently changed in the tsickle CLI (which almost nobody uses) is this:
https://github.com/angular/tsickle/commit/7d01aeefb2e168f0f55895b0f04b950e11f2ad2d

PS: @JaredNeil for any change, remember to make the commit message explain why you're doing it.

evmar:
Looking at this CL in particular I think it makes sense (to not fail compilations on warnings), so the only thing to wonder about is whether we want to warn users during their builds if they make tsickle mad.  The warnings are about stuff like "don't include types in your jsdoc".

JaredNeil:
We had some Typescript code that created a plugin interface object to interact with some old closure js code. It contained some closure annotations like this:
```
{
    'lucid':{
        'useragent': {
            /**
            * Determine if the user is using the chrome browser.
            * @memberOf    lucid.useragent
            * @method      is_chrome
            */
            'is_chrome': goog.bind(this.is_chrome, this),
        },
    }
}
```
And that's where we were getting the warnings.
However, since I opened this PR, we've actually done some refactoring and removed the closure comments, so this is no longer a problem for us. Happy to do some work if you still want this change, but I'm fine with closing the issue and PR as well.

alexeagle:
I'm inclined to leave the warnings fatal for now, because I feel that showing warnings in the stdout of a compiler just causes warning blindness, not fixing the warning. If we like it, we should have a linter for it that reports later in the development process.

evmar:
It sounds fine, but note that the consequence of this is that code written without being aware of tsickle may not compile under tsickle.  This can come up in an app that has some non-tsickle build and then they wanna run a tsickle-based build on the side.

PR #420, opened 2019-02-14
Support Angular plugin rollout: declare extra outs

alexeagle:
Angular ngtsc needs to emit two additional .js files for each .ts input to support legacy imports


alexeagle:
[sync CL](http://cl/234026312)

PR #423, opened 2019-02-18
fix(devserver): parent package static assets incorrectly injected

### LINKED ISSUE #409: devserver: asset injection of source files outside of current package not working
# 🐞 bug report

### Affected Rule

`ts_devserver`.

Seems to be Windows only. Something we need to fix after the windows fixes have
been merged. 

### Is this a regression?

Nope. This never seemed to work since the asset inlining has been introduced. It just works on non-windows due to the side-effect that `rules_go` also tries to resolve runfiles relatively to the working directory (for Bazel actions the PWD is by default the current workspace dir in the runfiles)

e.g. `my-server.runfiles/my-workspace = PWD`

### Description

Currently if a developer uses the newly introduced HTML asset inlining within
the `devserver` by doing something like that:

```bzl
ts_devserver(
  name = "server",
  index_html = "index.html",
  # Files outside the current package where this rule is defined break.
  # Files inside the current package work because the current package is 
  # a default root directory.
  static_files = ["//outside-of-current-package:my-file.css"]
)
```

This breaks inlining the `my-file.css` file because the path of the given
static file is relative to the _execroot_, while the devserver requires
manifest paths. 

e.g. inlined path is: `{host}/third_party/my-file.css` while it should be `{host}/{my_workspace_name}/third_party/my-file.css`.

This can be easily fixed by just providing the current workspace as root
directory in the `ts_devserver`.

devversion:
Currently whenever the `ts_devserver` tries to inject static
assets which are not part of the current Bazel package, but
part of the current workspace (e.g. file in a parent package),
incorrect asset URLs are injected.

This is because the `ts_devserver` currently uses the `File#path`
property which is relative to the execroot. So files which are not
part of the current Bazel package, but part of the execroot sources,
will be kept relative to the execroot. This is problematic because
the execroot is not part of the devserver root directories and
therefore the injected URLs are incorrect. Instead we should always
explicitly pass the `ts_devserver` implementation-specific manifest
paths to the asset injection function.

To be clear: This currently only affects Windows because if Bazel
symlinks the runfiles, the Go runfile helpers also look for the
requested files in the `working dirctory` (which is always the
the `devserver_runfiles/{workspace_name}` --> so runfiles within
the current workspace can be always resolved even though they
are not part of the root directory. This is something we need
to make consistent in the future, but is not related to that
specific bug (we should always precisely pass manifest paths)

Fixes #409.

### DIFF
--- a/internal/devserver/ts_devserver.bzl
+++ b/internal/devserver/ts_devserver.bzl
@@ -100,7 +100,7 @@ def _ts_devserver(ctx):
                 "/".join([ctx.bin_dir.path, ctx.label.package]),
                 "/".join([ctx.genfiles_dir.path, ctx.label.package]),
             ],
-            [f.path for f in ctx.files.static_files] + [bundle_script],
+            [_short_path_to_manifest_path(ctx, f.short_path) for f in ctx.files.static_files] + [bundle_script],
             injected_index,
         )
         devserver_runfiles += [injected_index]
--- a/internal/e2e/npm_packages/ts_devserver/package-template.json
+++ b/internal/e2e/npm_packages/ts_devserver/package-template.json
@@ -10,6 +10,8 @@
   },
   "scripts": {
     "pretest": "webdriver-manager update $CHROMEDRIVER_VERSION_ARG",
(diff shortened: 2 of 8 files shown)


alexeagle:
Mailed cl/234675750 to @kyliau

PR #426, opened 2019-02-20
Allow configuration devmode language level in ts_library

gregmagolan:


internal/tsc_wrapped/tsconfig.ts
```diff
@@ -36,7 +36,10 @@ export interface BazelOptions {
   /** If true, convert require()s into goog.module(). */
   googmodule: boolean;
 
-  /** If true, emit ES5 into filename.es5.js. */
+  /**
+   * If true, emit devmode output into filename.js.
+   * If false, emit prodmode output into filename.closure.js.
+   */
   es5Mode: boolean;
```
gregmagolan: Badly named. Can we rename this to devmode without breaking g3?

PR #429, opened 2019-02-22
Tear out the OSS-only parts of rules_typescript

alexeagle:
These can now live in rules_nodejs which has a source dependency on this repo.


DEVELOPING.md
```diff
@@ -48,23 +48,3 @@ an `@npm` workspace with npm dependencies.
 
 Note, with this workflow the downstream version of `@npm//typescript` will be used to compile the `ts_library` targets in `npm_bazel_typescript`.
```
gregmagolan: The bit about testing downstream can be removed as well. It belongs in the rules_nodejs repo and should be generalized for any @bazel/foobar package.

WORKSPACE
```diff
@@ -12,24 +12,13 @@
...
 
 rules_typescript_dev_dependencies()
 
-# Load rules_karma dependencies
-load("@npm_bazel_karma//:package.bzl", "rules_karma_dependencies")
-
-rules_karma_dependencies()
-
 # Setup nodejs toolchain
 load("@build_bazel_rules_nodejs//:defs.bzl", "node_repositories", "yarn_install")
```
gregmagolan: nit: node_repositories() not needed :)

DEVELOPING.md
alexeagle: okay, removing this whole file

PR #442, opened 2019-04-23
Don't test on Ubuntu 14.04

philwo:
Ubuntu 14.04 is about to be end-of-life and Bazel CI will stop supporting it shortly afterwards.

Context: https://groups.google.com/d/msg/bazel-dev/_D6XzfNkQQE/8TNKiNmsCAAJ

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,16 +1,5 @@
 ---
 platforms:
-  ubuntu1404:
-    run_targets:
-    - "//:gazelle"
-    - "@nodejs//:yarn"
-    build_targets:
-    - "..."
-    test_flags:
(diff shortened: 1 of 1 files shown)


.bazelci/presubmit.yml
```diff
@@ -1,24 +1,24 @@
...
-  ubuntu1604:
+  ubuntu1804:
     run_targets:
     - "//:gazelle"
     - "@nodejs//:yarn"
     build_targets:
     - "..."
     test_flags:
-    # TODO(gregmagolan): shared libs needed by chrome & firefox not available on ubuntu1604
+    # TODO(gregmagolan): shared libs needed by chrome & firefox not available on ubuntu1404
```
alexeagle: should be 1804?

PR #446, opened 2019-05-09
Update to rules nodejs 0.29.0 & bazel 0.25.1

gregmagolan:


package.json
```diff
@@ -7,10 +7,10 @@
         "semver": "5.6.0",
         "source-map-support": "0.5.9",
         "tsutils": "2.27.2",
-        "@bazel/bazel": "~0.22.0",
+        "@bazel/bazel": "0.25.1",
         "@bazel/buildifier": "^0.20.0",
         "@bazel/ibazel": "^0.2.0",
```
kyliau: on a side note: this version of ibazel is very old, we should probably upgrade it to 0.10.2

gregmagolan: Maybe we should remove both bazel & ibazel since they are not used by this repo in CI and this is just a sync point now?

gregmagolan: Maybe we should remove both bazel & ibazel since they are not used by this repo in CI and this is just a sync point now?

Issue #443, opened 2019-04-25
incompatible_depset_is_not_iterable


laurentlb:
According to Bazel CI, this command fails: `bazel build... --incompatible_depset_is_not_iterable`

```
File "/home/bazel/.cache/bazel/_bazel_bazel/664e1f005eb545012940353990310184/external/npm_bazel_typescript/internal/common/tsconfig.bzl", line 278, in create_tsconfig
--
  | [((workspace_path + "/") + f.path) for f in files]
  | type 'depset' is not iterable. Use the `to_list()` method to get a list. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check.
```

Can you take a look at this?

laurentlb:
I think you need to update the dependency on rules_nodejs.

alexeagle:
need to cut a rules_nodejs release to get that fix...

alexeagle:
Okay should be fixed by 09f858aec16969660092cd83d62807247ab11f6e

laurentlb:
No, there's still an error:

```
$ bazel build --incompatible_depset_is_not_iterable ... --nobuild
	File "/.../96315885d9b5519569de96c9f8ab613a/external/npm_bazel_typescript/internal/common/tsconfig.bzl", line 278, in create_tsconfig
		[((workspace_path + "/") + f.path) for f in files]
type 'depset' is not iterable. Use the `to_list()` method to get a list. Use --incompatible_depset_is_not_iterable=false to temporarily disable this check.
```

(cc @vladmos)

laurentlb:
Since tsconfig.bzl was changed, is it just a matter of updating inter-dependencies between rules_nodejs and rules_typescript, @alexeagle?

PR #451, opened 2019-05-21
Fixes for --incompatible_depset_is_not_iterable

alexeagle:
Sadly because of the sandwich of rules_nodejs calling rules_typescript which calls back into a function provided by rules_nodejs, this takes a couple releases to unwind.
@bazel/typescript will have to be released again with this fix before we can turn on the option.

See #443


internal/common/tsconfig.bzl
```diff
@@ -54,6 +54,9 @@ def create_tsconfig(
     Returns:
       A nested dict that corresponds to a tsconfig.json structure
     """
+    if (type(files) != type(depset())):
+      fail("expected files argument to be a depset, got", type(files))
```
kyliau: nit: the second parameter of fail() is supposed to be the attribute that caused the error:
https://docs.bazel.build/versions/master/skylark/lib/globals.html#fail

PR #454, opened 2019-05-31
Opt-in to --incompatible_depset_is_not_iterable

alexeagle:


package.json
```diff
@@ -3,11 +3,11 @@
     "homepage": "https://github.com/bazelbuild/rules_typescript",
     "license": "Apache-2.0",
     "devDependencies": {
-        "@bazel/bazel": "0.25.1",
+        "@bazel/bazel": "^0.26.0",
         "@bazel/buildifier": "^0.20.0",
         "@bazel/ibazel": "^0.2.0",
-        "@bazel/jasmine": "^0.29.0",
-        "@bazel/typescript": "0.29.0",
+        "@bazel/jasmine": "^0.30.0",
```
kyliau: In the rules nodejs repo, `@bazel/*` packages are pinned to a specific version instead of tilde/caret ranges. 
Should the same be done here?

alexeagle: that's just an artifact of how renovate is configured by default.

We are okay with the lock file ensuring that our version don't float, and these version strings aren't used directly by users so we don't have to mitigate users forgetting to use lockfiles themselves (which is why we avoid ranges in the CLI dependencies)

alexeagle:
Thanks Greg!

PR #455, opened 2019-06-02
Fix path to worker_protocol.proto in ts_library for @npm/typescript package...

gregmagolan:
…downstream in rules_nodejs

This was broken after rules_nodejs was updated set `--preserve-symlinks-main` for 0.30.2 and rules_typescript was updated to rules_nodejs 0.30.2 here https://github.com/bazelbuild/rules_typescript/commit/c6fdfefe4b245683de4836a38bbc0ca4bc4ae8eb which made the following change necessary to fix the local build https://github.com/bazelbuild/rules_typescript/commit/c6fdfefe4b245683de4836a38bbc0ca4bc4ae8eb#diff-118aceecde6f4c3cc18c0811d4bb3f48.

That change fixed the local build but broke the downstream usage of ts_library since the path to the proto file in the `@bazel/typescript` npm package needs to be relative.

### DIFF
--- a/internal/tsc_wrapped/worker.ts
+++ b/internal/tsc_wrapped/worker.ts
@@ -72,7 +72,7 @@ declare namespace workerProto {
  */
 function loadWorkerPb() {
   const protoPath =
-      '../worker_protocol.proto';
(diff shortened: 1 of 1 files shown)


alexeagle:
merged in 4ae7cda81c7ea3215a024f10b00ee38a0755549d

PR #458, opened 2019-06-19
Update to rules_nodejs 0.32.0...

gregmagolan:
* `@npm//node_modules/foobar:foobar.js` labels changed to `@npm//:node_modules/foobar/foobar.js` with fix for https://github.com/bazelbuild/rules_nodejs/issues/802.

* tsconfig.ts now merge compilerOptions plugins @bazel/tsetse disabledRules from tsconfig extends which unblocks https://github.com/angular/angular/pull/31019


package.bzl
```diff
@@ -30,12 +30,8 @@ def rules_typescript_dev_dependencies():
     _maybe(
         http_archive,
         name = "build_bazel_rules_nodejs",
-        # This file just copy-pasted from output of `git diff`
-        patches = ["//:rules_nodejs_pr756.patch"],
-        # According to docs, this is needed for patches generated by git
-        patch_args = ["-p1"],
-        sha256 = "e04a82a72146bfbca2d0575947daa60fda1878c8d3a3afe868a8ec39a6b968bb",
-        urls = ["https://github.com/bazelbuild/rules_nodejs/releases/download/0.31.1/rules_nodejs-0.31.1.tar.gz"],
+        sha256 = "06cb04f4f745e37d542ec6883a2896029715a591c0e44c5d250a268d3752f865",
```
alexeagle: need to delete the patch file too

internal/tsc_wrapped/tsconfig.ts
```diff
@@ -276,6 +277,38 @@ export function parseTsconfig(
           devmodeTargetOverride: isUndefined(existingBazelOpts.devmodeTargetOverride)
             ? newBazelBazelOpts.devmodeTargetOverride
             : existingBazelOpts.devmodeTargetOverride,
+        };
+
+        // merge compilerOptions plugins @bazel/tsetse disabledRules
+        const existingCompilerOptions: ts.CompilerOptions = existingConfig.compilerOptions || {};
```
alexeagle: we need a unit test for this file...

package.bzl
gregmagolan: 👍

internal/tsc_wrapped/tsconfig.ts
alexeagle: I added this:
```
  it('honors plugins config in users tsconfig', () => {
    const userTsconfig = {
      compilerOptions: {
        plugins: [{
          name: '@bazel/tsetse',
          disabledRules: ['must-use-promises'],
        }]
      }
    };
    const generatedTsconfig = {
      extends: './user.tsconfig',
      files: ['a.ts'],
      compilerOptions: {},
    };
    const files = {
      [resolveNormalizedPath('path/to/user.tsconfig.json')]:
          JSON.stringify(userTsconfig),
      [resolveNormalizedPath('path/to/generated.tsconfig.json')]:
          JSON.stringify(generatedTsconfig),
    };
    const host = mockHost(files);
    const [parsed, diagnostics, {target}] =
        parseTsconfig('path/to/generated.tsconfig.json', host);
    expect(diagnostics).toBeNull();
    if (!parsed) {
      fail('Expected parsed');
    } else {
      const {options, bazelOpts, files, config} = parsed;
      if (!options) {
        fail('no options');
      } else {
        const plugins = options['plugins'] as PluginImportWithConfig[] || [];
        expect(plugins.length).toBe(1);
        const [tsetsePlugin] = plugins;
        expect(tsetsePlugin.name).toBe('@bazel/tsetse');
        const disabledRules = tsetsePlugin['disabledRules'] as Array<string>;
        expect(disabledRules.length).toBe(1);
        expect(disabledRules[0]).toBe('must-use-promises');
      }
    }
  });
```
but it passes without your change.

Does it need to be 
generatedTsconfig extends userA extends userB
with userB defining the disabledRules?

That doesn't seem to be what happens in https://github.com/angular/angular/pull/31019/files#diff-36fb2cb41cec12240538690af845c47b so I'm confused

alexeagle:
I'll merge the update part first, then the feature for tsconfig extends tsetse disabledRules

alexeagle:
seems like we don't need the complexity of overriding the tsetse disabledRules

PR #489, opened 2020-02-09
Add external_module_roots arg to create_tsconfig

volkangurel:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)
- [x] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: https://github.com/bazelbuild/rules_nodejs/issues/1033

## What is the new behavior?
The `create_tsconfig` function now has an `external_module_roots` arg so that the `ts_library` rule can insert explicit paths into `tsconfig.json` for scoped packages. The change for `ts_library` will be made as a separate PR once this change merges in (a potential change is like so https://github.com/volkangurel/rules_nodejs/commit/04420072a11a4cbbe69e3c0f3c6c25566025def8)

- [ ] Yes
- [x] No

## Other information

N/A

### DIFF
--- a/internal/common/tsconfig.bzl
+++ b/internal/common/tsconfig.bzl
@@ -17,7 +17,6 @@
 
 load(":common/module_mappings.bzl", "get_module_mappings")
 
-
 _DEBUG = False
 
 def create_tsconfig(
@@ -33,6 +32,7 @@ def create_tsconfig(
         extra_root_dirs = [],
         module_path_prefixes = None,
         module_roots = None,
+        external_module_roots = None,
         node_modules_root = None):
     """Creates an object representing the TypeScript configuration to run the compiler under Bazel.
 
@@ -50,6 +50,7 @@ def create_tsconfig(
       extra_root_dirs: Extra root dirs to be passed to tsc_wrapped.
       module_path_prefixes: additional locations to resolve modules
       module_roots: standard locations to resolve modules
+      external_module_roots: external locations to resolve modules
       node_modules_root: the node_modules root path
 
     Returns:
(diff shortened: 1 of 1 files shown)


volkangurel:
@googlebot I signed it!

alexeagle:
Hi, thanks Volkan!

My understanding of the associated issue is that this is an issue in TypeScript - they only understand module names that come from the automatic resolver.

For this change, I would have expected that the aspect we run to collect the module mappings should find all the external modules, so this workaround feels like it might be happening in the wrong layer. It seems like the BUILD file we generate for scoped typings packages should maybe expose the module_name and module_root so it happens to get picked up using the existing logic.

Do you have a small repro of the current issue?

(Also as a note, @gregmagolan and I have been working on a new implementation for ts_library that doesn't do any of this path manipulation, by making sure TS sees a node_modules directory that matches its expectations. But no estimate on when we might finish, depends on how we can get paid for it)

volkangurel:
Hey Alex!

I have a small repro of the issue here: https://github.com/volkangurel/bazel-scoped-package-example. The first commit in the repo is the vanilla `@bazel/create` and the second commit is a minimal set of changes to (fail to) build a typescript file that imports a scoped package.

To be honest I'm not very familiar with how BUILD files are generated for packages, but that sounds like a better place to implement this change for sure. Can you point me in the right direction on how I can go about developing something that will expose the module_name and module_root from these BUILD files? Is there an example of doing something similar somewhere?

Also looking forward to that new implementation you're thinking about!

PR #490, opened 2020-02-11
Switch tsc_wrapped to a .js entry point

alexeagle:
This lets it work under the linker, where the execpath to the .js file
is hard to calculate from the .ts input file path.


.bazelrc
```diff
@@ -1,4 +1,4 @@
-build --strategy=TypeScriptCompile=worker
+build --strategy=TypeScriptCompile=worker --worker_sandboxing
```
gregmagolan: why the `--worker_sandboxing`?

alexeagle: I was using that to help debug. But then it caused an error on Windows
```
---8<---8<--- Start of log, file at C:/b/mp6jso7u/bazel-workers/worker-2-TypeScriptCompile.log ---8<---8<---
ERROR: Manifest file C:\b\mp6jso7u\bazel-workers\worker-2-TypeScriptCompile\build_bazel_rules_typescript\bazel-out\host\bin\external\npm\@bazel\typescript\bin\tsc_wrapped.bat.runfiles_manifest does not exist.
---8<---8<--- End of log ---8<---8<---
(21:13:58) ERROR: C:/b/bk-windows-msd2/bazel/google-typescript-presubmit/internal/tsetse/tests/ban_promise_as_condition/BUILD:42:1: Couldn't build file internal/tsetse/tests/ban_promise_as_condition/negatives.js: Compiling TypeScript (devmode) //internal/tsetse/tests/ban_promise_as_condition:negatives failed: Worker process quit or closed its stdin stream when we tried to send a WorkRequest:
```
so I reverted that line in the g3 import of this PR

PR #491, opened 2020-02-14
Fix worker loading when using require.resolve

alexeagle:
The breakage was observed downstream in @bazel/worker package which no longer worked.

g3 instructions: There is a copybara transform in third_party/bazel_rules/rules_typescript/copy.bara.sky
which already transforms the protoPath variable to the prefix '../..'
this needs to be changed to '.' so that the output of copybara looks like this PR.
See the README.google.md in the rules_typescript directory for hints about checking that.

### DIFF
--- a/internal/tsc_wrapped/worker.ts
+++ b/internal/tsc_wrapped/worker.ts
@@ -72,7 +72,7 @@ declare namespace workerProto {
  */
 function loadWorkerPb() {
   const protoPath =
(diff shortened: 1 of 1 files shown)


alexeagle:
@soldair would you like to merge this one?

soldair:
It seems like to don't have the required permissions here. it's good we found this now =)

PR #492, opened 2020-03-04
feat: add devmodeModuleOverride to tsconfig bazelOptions

gregmagolan:
By default devmode module format is UMD (or CommonJS if googmodule is set). This option allows that to be overridden (if googmodule is not set).

This feature is requested by @filipesilva for the angular-cli bazel migration.

PR in rules_nodejs adds test coverage for the new tsconfig setting: https://github.com/bazelbuild/rules_nodejs/pull/1687

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/tsc_wrapped/tsconfig.ts
+++ b/internal/tsc_wrapped/tsconfig.ts
@@ -201,6 +201,17 @@ export interface BazelOptions {
    * future.
    */
   devmodeTargetOverride?: string;
+
+  /**
+   * Override for module format to use for devmode.
+   *
+   * This setting can be set in a user's tsconfig to override the default
+   * devmode module.
+   *
+   * EXPERIMENTAL: This setting is experimental and may be removed in the
+   * future.
+   */
+  devmodeModuleOverride?: string;
 }
 
 export interface ParsedTsConfig {
@@ -285,6 +296,10 @@ export function parseTsconfig(
           devmodeTargetOverride: isUndefined(existingBazelOpts.devmodeTargetOverride)
             ? newBazelBazelOpts.devmodeTargetOverride
             : existingBazelOpts.devmodeTargetOverride,
+
+          devmodeModuleOverride: isUndefined(existingBazelOpts.devmodeModuleOverride)
+            ? newBazelBazelOpts.devmodeModuleOverride
+            : existingBazelOpts.devmodeModuleOverride,
         }
       }
 
@@ -346,7 +361,7 @@ export function parseTsconfig(
         break;
       default:
         console.error(
-            'WARNING: your tsconfig.json file specifies an invalid bazelOptions.devmodeTargetOverride value of: \'${bazelOpts.devmodeTargetOverride\'');
+            `WARNING: your tsconfig.json file specifies an invalid bazelOptions.devmodeTargetOverride value of: '${bazelOpts.devmodeTargetOverride}'`);
     }
   }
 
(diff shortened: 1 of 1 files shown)


evmar:
Can you motivate the change more?  Why do we need more than two output formats?  Can we instead replace one of the outputs with whichever one you want?

gregmagolan:
> Can you motivate the change more? Why do we need more than two output formats? Can we instead replace one of the outputs with whichever one you want?

A default of UMD makes sense for many cases in the OSS world as the devmode module format needs to be the default UMD to work with ts_devserver & karma_web_test as these both use concatjs and require a named module format. However, there are other cases that users would like to use ts_library and get a non-UMD devmode output.

In the angular-cli case the devmode output .js files are put into their release packages which currently have CJS in them and can't be changed to UMD due to a typescript compilation issue with UMD. In other cases, users are not using ts_devserver or karma_web_test and would prefer a less verbose module format in their devmode .js outputs. Prodmode has to stay esm as the outputs are named .mjs.

filipesilva:
The typescript bug is https://github.com/microsoft/TypeScript/issues/36780 and it seems to affect other usage patterns in UMD. Aside from this we want to ship `commonjs` anyway because we are building node libraries exclusively.

I can imagine an alternative setup where instead of outputting `commonjs` from `ts_library` we instead output `.mjs` and then transpile that to `commonjs`. But that seems roundabout. We'd need another set of rules everywhere for it, and those tools would have to use something that can actually output `commonjs`, since wouldn't be available with `ts_library`.

Toxicable:
We have a use case for this where we have libs that we want to consume with the angular architect bazel rule.
When consuming UMD format via architect / build-angular it complines as an errors or wanring since UMD has dynamic require statements.

evmar:
Would it make sense to just obey the tsconfig's module setting then?

filipesilva:
That's how I thought it originally worked. I imagine there's a reason against it, but I don't know what it is.

Toxicable:
I'd propose that it read the tsconfig and if no value is set use `amd` otherwise use the value set.

Howevery this is a breaking change, on the current behaviour and we couldn't release till 2.0.
To amend that we could add a attribute to `ts_library` that mirriors the `module` option in the tsconfig, which will do the same as `devmodeModuleOverride` but mark it to be removed in 2.0 thus making the brekaing change in 2.0 but allowing this feature into 1.x

evmar:
I think you could also do it with an 'obeyTsConfigModuleSetting' flag in that case.

Toxicable:
Ah yes, that would be a better flag, default being false to avoid break changes then consider flipping to true in 2.0

gregmagolan:
Circled back to this. An `obeyTsConfigModuleSetting` is non-trivial since the tsconfig used by ts_library is generated in starlark (tsconfig.bzl) and that is where the forcing of `umd` or `esnext` happens:
```
        # The "typescript.es5_sources" provider is expected to work
        # in both nodejs and in browsers, so we use umd in devmode.
        # NOTE: tsc-wrapped will always name the enclosed AMD modules
        # For production mode, we leave the module syntax alone and let the
        # bundler handle it (including dynamic import).
        # Note, in google3 we override this option with "commonjs" since Tsickle
        # will convert that to goog.module syntax.
        "module": "umd" if devmode_manifest or has_node_runtime else "esnext",
```
At that point it is not possible to know if `obeyTsConfigModuleSetting` is in the user's tsconfig in order to not force the module format to either `umd` or `esnext`. The same problems exists for the language level which is why we have `devmodeTargetOverride` instead of `obeyTsConfigTargetSettings`.

I propose we land the `devmodeModuleOverride` as it is in line with `devmodeTargetOverride` which already exists.

Toxicable:
@gregmagolan SGTM

evmar:
Since this is a build-level concern, why not specify it in the build file?

gregmagolan:
> Since this is a build-level concern, why not specify it in the build file?

Its possible and easier to setup for the oss ts_library since we can make that change in rules_nodejs; tho I don't have a strong argument as to why this setting is different from the other `bazelOpts` that can be specified and then we'd be in the unusual spot of having a `devmodeTargetOverride` bazelOpt and a `devmode_module_override` ts_library attribute.

Is devmodeTargetOverride override used at all in google3? If not, we could remove that and add `devmode_target_override` to ts_library in rules_nodejs and convert angular over to the attribute fairly easily.

evmar:
Sorry for all my questions.  devmodeTargetOverride is not used in Google.  (We don't let users control tsconfig at all.)  I'm not sure I'm the right reviewer for this in general, I will try to find someone better.

PS here is some thinking around adding options like these that I've found useful: http://neugierig.org/software/blog/2018/07/options.html

gregmagolan:
> Sorry for all my questions. devmodeTargetOverride is not used in Google. (We don't let users control tsconfig at all.) I'm not sure I'm the right reviewer for this in general, I will try to find someone better.
> 
> PS here is some thinking around adding options like these that I've found useful: http://neugierig.org/software/blog/2018/07/options.html

Thanks for the link @evmar and the questions are appreciated. You brought up a good point about ts_config bazelOpts vs rule attributes. I'll give it some thought and get some feedback on that from the angular team.

gregmagolan:
Went with https://github.com/bazelbuild/rules_nodejs/pull/1687 instead which can be handled entirely in rules_nodejs and seems like a more principled solution.

PR #493, opened 2020-03-04
Allow all files contained in allowed directory for strictDeps

CaerusKaru:
Currently, if you pass a directory without explicitly
enumerating all files in the directory, the strictDeps rule
will fail. This checks the path of any directories that are
allowed to see if a given files is contained to determine if
it really is transitive or not.

In essence, allow dir/** pattern for strictDeps

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature (please, look at the "Scope of the project" section in the README.md file)
- [ ] Build related changes
- [ ] CI related changes

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/internal/tsc_wrapped/strict_deps.ts
+++ b/internal/tsc_wrapped/strict_deps.ts
@@ -30,6 +30,9 @@ export interface StrictDepsPluginConfig {
 /** The TypeScript diagnostic code for "Cannot find module ...". */
 export const TS_ERR_CANNOT_FIND_MODULE = 2307;
 
+/** RegExp used to test whether a file is a descendent of a parent directory */
+const PATH_TEST_REGEX = /^(\.{1,2}\/)*(\.{1,2})?$/;
+
 /**
  * The strict_deps plugin checks the imports of the compiled modules.
  *
@@ -68,8 +71,15 @@ export function checkModuleDeps(
   function stripExt(fn: string) {
     return fn.replace(/(\.d)?\.tsx?$/, '');
   }
-  const allowedMap: {[fileName: string]: boolean} = {};
-  for (const d of allowedDeps) allowedMap[stripExt(d)] = true;
+  const allowedFiles: Set<string> = new Set();
+  const allowedDirs: Set<string> = new Set();
+  for (const d of allowedDeps) {
+    const path = stripExt(d);
+    allowedFiles.add(path);
+    if (path === d) {
+      allowedDirs.add(path);
+    }
+  }
 
   const result: ts.Diagnostic[] = [];
   for (const stmt of sf.statements) {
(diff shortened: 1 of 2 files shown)


DavidANeil:
Wouldn't this cause false positives in the case of
`src/a/BUILD` -> `src/a/a/BUILD`
`src/a/b/BUILD` -> `src/a/BUILD`

And then a dep `src/a/b/f.ts` -> `src/a/b/f.ts` would be allowed, even though the dep isn't specified?

CaerusKaru:
@DavidANeil Can you elaborate on your example a little bit? (Also not sure but I think you might've made a typo above)

internal/tsc_wrapped/strict_deps.ts
```diff
@@ -87,8 +97,20 @@ export function checkModuleDeps(
     }
     const declFileNames =
         sym.declarations.map(decl => decl.getSourceFile().fileName);
-    if (declFileNames.find(
-            declFileName => !!allowedMap[stripExt(declFileName)])) {
+    const checkDir = (filename: string): boolean => {
+      for (const dir of allowedDirs) {
+        const descendent = PATH_TEST_REGEX.test(path.posix.relative(filename, dir));
```
alexeagle: the regex seems kinda overkill, don't you just need `.startsWith('../')`?

internal/tsc_wrapped/strict_deps.ts
```diff
@@ -68,8 +71,15 @@ export function checkModuleDeps(
   function stripExt(fn: string) {
     return fn.replace(/(\.d)?\.tsx?$/, '');
   }
-  const allowedMap: {[fileName: string]: boolean} = {};
-  for (const d of allowedDeps) allowedMap[stripExt(d)] = true;
+  const allowedFiles: Set<string> = new Set();
+  const allowedDirs: Set<string> = new Set();
+  for (const d of allowedDeps) {
+    const path = stripExt(d);
```
alexeagle: this assumes that a file has an extension and a directory doesn't? feels like the wrong condition.

alexeagle (commented):
could you update the commit message to explain that this is adding support for a TreeArtifact that appears in the srcs of ts_library? (that condition doesn't happen internal at Google AFAIK)

internal/tsc_wrapped/strict_deps.ts
CaerusKaru: This assumes that a "directory" is anything that doesn't end in `.ts` or `.d.ts` or `.tsx` (stripExt should be stripTsExt IMO). I think this is OK because the rule would treat those as files anyway.

internal/tsc_wrapped/strict_deps.ts
CaerusKaru: No because then the rule might descend into a parent. E.g. `../../../parent/no/don't/go/deeper`. Technically all files are connected to each other in some form or another if you go high enough and then descend back down. We want to make sure the file is exclusively a descendant of the allowed directory.

PR #501, opened 2020-07-16
chore: update to rules_go 0.22.4 & rules_nodejs 2.0.0-rc.2

gregmagolan:


gregmagolan:
Seems we can't upgrade to a newer version of rules_go because of a runfiles issue which we patch for. The patch no longer applies cleanly to the latest version.
```
        patch_args = ["-p1"],
        # Patch out a breaking change to runfiles support library
        # See discussion on https://github.com/bazelbuild/rules_go/pull/2076
        patches = ["@build_bazel_rules_typescript//:revert_rules_go_commit_4442d82a001f378d0605cbbca3fb529979a1c3a6.patch"],
```

alexeagle:
Yeah merge.conflict in rules_go... Sorry 🙄

gregmagolan:
Went down this rabbit hole as I wanted to bring in rules_docker to the root rules_nodejs WORKSPACE for a test but rules_docker requires the latest rules_go. Dependency spaghetti. I ended up putting my test in a nested e2e workspace instead.

gregmagolan:
It should also be noted that rules_go https://github.com/bazelbuild/rules_go/releases/tag/v0.23.0 "migrated to Starlark build settings and configuration transitions" which changes where non-default configuration files are outputted which breaks the rules_typescript npm package. The go binaries end up in paths such as `bazel-out/darwin-fastbuild-ST-05b09dfdd2e5ae92d24745e4e699444f165b36fe4c5cff40bcd8e02b7f69ac79/bin/external/build_bazel_rules_typescript/devserver/devserver-linux_x64` instead of the old location under `bazel-out/darwin-fastbuild` and pkg_npm doesn't account for the folder as it only handles the default bin folder `bazel-out/darwin-fastbuild`

PR #503, opened 2020-08-03
Remove deprecated flag from bazelrc

Wyverald:
--experimental_allow_incremental_repository_updates has been deleted in https://github.com/bazelbuild/bazel/commit/fdf10c34c82c563fb7fc7f262de32e1f7af59e2d

### DIFF
--- a/.bazelrc
+++ b/.bazelrc
@@ -7,6 +7,3 @@ test:debug --test_arg=--node_options=--inspect-brk --test_output=streamed --test
 # Turn off legacy external runfiles
 run --nolegacy_external_runfiles
 test --nolegacy_external_runfiles
-
-# Enable the "Managed Directories" feature
-common --experimental_allow_incremental_repository_updates


Wyverald:
poke @kyliau @mgechev

c-parsons:
ping

kyliau:
https://github.com/bazelbuild/bazel/commit/fdf10c34c82c563fb7fc7f262de32e1f7af59e2d has only landed on master, but not released right? Shall we remove this after upgrading the version?

aiuto:
Yes, you should wait for the update. It looks like:
The flag default was made true on July 17, in fc057437ec054e8f154e97ef87eaeabcefed53bf
The flag was remove on July 31, in fdf10c34c82c563fb7fc7f262de32e1f7af59e2d

That would mean it exists and defaults false false in Bazel 3.4 and will not exist at all in Bazel 3.5.
We should have made that window bigger.

aiuto:
OTOH.... since you make releases, you might as well flip now and then you will work with Bazel at head and the bazel 3.5rc candidate. You are currently broken there. https://buildkite.com/bazel/bazel-at-head-plus-downstream/builds/1608#ceace0eb-4175-4b53-8717-71e722856fe6

comius:
friendly ping

aiuto:
FWIW.  Bazel 3.5 rc4 is ready to go as the real release. We're just waiting on the blog post review.

laurentlb:
@kyliau, it looks like this can be merged now?

kyliau (approved):

laurentlb:
Thanks! Could you merge too?

PR #511, opened 2020-09-21
fix: remove incompatible license attribute from compilation

devversion:
In the past, the `license` attribute has been added. This attribute seems to be no longer available in the latest versions of Bazel. The attribute is only available with the "no-incompatible" Bazel command line flag. See: https://cs.opensource.google/bazel/bazel/+/master:src/main/java/com/google/devtools/build/lib/packages/semantics/BuildLanguageOptions.java;l=395-405?q=incompatible_no_attr_license&ss=bazel.

We should make sure the rules are compatible with the most recent versions of Bazel.

### DIFF
--- a/.bazelci/presubmit.yml
+++ b/.bazelci/presubmit.yml
@@ -1,5 +1,5 @@
 ---
-bazel: 2.1.0
+bazel: 3.0.0
 platforms:
   ubuntu1604:
     run_targets:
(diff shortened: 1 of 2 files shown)


alexeagle:
Included in https://github.com/bazelbuild/rules_nodejs/pull/2237 - you can probably close this one

